PR #19, opened 2017-02-09
Headers

smccauliff:


src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -251,6 +288,34 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
+   */
+  public static ProducerRecord<byte[], byte[]> serializeWithHeaders(ExtensibleProducerRecord<byte[], byte[]> xRecord,
+    HeaderSerializerDeserializer headerParser) {
+
+    if (!xRecord.hasHeaders()) {
+      return new ProducerRecord<byte[], byte[]>(xRecord.topic(), xRecord.partition(), xRecord.timestamp(), xRecord.key(), xRecord.value());
+    }
+
+    int headersSize = headerParser.serializedHeaderSize(xRecord.headers());
+    int serializedValueSize = (xRecord.value() == null ? 0 : xRecord.value().length) + headersSize;
```
becketqin: valueSizeWithHeaders?

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -342,4 +399,14 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
     }
   }
 
+  private final class SimplePartitionerImpl implements SimplePartitioner {
```
becketqin: I am actually not sure if this is necessary. Usually we want to avoid having something very similar to the existing configurations. Ideally we want to just use the open source partitioner. I understand that we do not have easy access to the Cluster which is required by the open source Kafka partitioner interface, maybe we can pass in null in this case. If there is no good solution, I would rather remove this partitioner at this point. It is always much easier to add something than remove some existing things.

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializer.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ *  <pre>
+ *  8 bytes magic number
+ *  1 byte version/flags
+ *  4 byte length of all headers
+ *  Varies Header key and length fields are encoded in a 4TLV format using network byte ordering.
+ *  </pre>
+ */
+public class DefaultHeaderSerializerDeserializer implements HeaderSerializerDeserializer {
```
becketqin: As mentioned in the other comment. I think we should have separate interface for serializer and deserializer. This class implement both.

src/test/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerIntegrationTest.java
```diff
@@ -62,6 +73,8 @@
  */
 public class LiKafkaConsumerIntegrationTest extends AbstractKafkaClientsIntegrationTestHarness {
 
+  private static final Logger LOG = LoggerFactory.getLogger(LiKafkaConsumerIntegrationTest.class);
```
becketqin: It seems typically we do not use log4j in the unit tests.

src/test/java/com/linkedin/kafka/clients/producer/DocumetationTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+
+/**
+ * This is a simple code example we use in the documentation.
+ */
+public class DocumetationTest extends AbstractKafkaClientsIntegrationTestHarness {
```
becketqin: Maybe BasicHeadertest?

src/test/java/com/linkedin/kafka/clients/producer/LiKafkaProducerIntegrationTest.java
```diff
@@ -78,8 +87,197 @@ public void testSend() throws IOException, InterruptedException {
...
+    consumer.subscribe(Collections.singleton(topic));
+
+    int messageCount = 0;
+    long startMs = System.currentTimeMillis();
+    while (messageCount < 4 && System.currentTimeMillis() < startMs + 30000) {
+
+      ConsumerRecords<String, String> consumerRecords = consumer.poll(10);
+      for (ConsumerRecord<String, String> consumerRecord : consumerRecords) {
+        ExtensibleConsumerRecord<String, String> xRecord = (ExtensibleConsumerRecord<String, String>) consumerRecord;
+        switch (messageCount) {
```
becketqin: Should we check the message offset as well?

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+public class DefaultHeaderSerializerDeserializerTest {
+  private static final char UNICODE_REPLACEMENT_CHARACTER = 65533;
+
+  @Test
+  public void invalidUtf8Magic() throws Exception {
```
becketqin: invalidUtf8MagicTest?

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.EOFException;
+import java.util.Arrays;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+public class DefaultHeaderSerializerDeserializerTest {
```
becketqin: Should we have a basic serde test?

src/main/java/com/linkedin/kafka/clients/consumer/ExtensibleConsumerRecord.java
```diff
@@ -0,0 +1,100 @@
...
+
+    if (headers == null) {
+      headers = new LazyHeaderListMap();
+    }
+
+    headers.put(headerKey, value);
+  }
+
+
+  public Set<Integer> headerKeys() {
```
smccauliff: I will update the javadocs for public methods.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -118,46 +112,43 @@ private LiKafkaConsumerImpl(LiKafkaConsumerConfig configs,
...
+    _keyDeserializer.configure(configs.originals(), true);
+    _valueDeserializer = valueDeserializer != null ? valueDeserializer :
         configs.getConfiguredInstance(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, Deserializer.class);
-    vDeserializer.configure(configs.originals(), false);
+    _valueDeserializer.configure(configs.originals(), false);
 
     // Instantiate consumer record processor
-    _consumerRecordsProcessor = new ConsumerRecordsProcessor<>(assembler, kDeserializer, vDeserializer,
-                                                               messageOffsetTracker, auditor);
+    _consumerRecordsProcessor = new ConsumerRecordsProcessor(assembler, messageOffsetTracker);
```
smccauliff: I would rather hide the ConsumerRecordsProcessor and not let people reuse it.  That being said resuse is even better when the CRP just does one thing like assemble large messages (or pass messages through) and do the necessary offset tracking.  If we (or someone else) has their own clients and has plugins (or something like them) then they can integrate the CRP into their record processing flow where it is needed.  In our case we may want to have plugins that run before or after deserialization and before or after large message support.  Having large message also do audit and deserialization conflates the various purposes of the CRP.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -257,10 +248,78 @@ public void unsubscribe() {
           }
         }
       }
-      processedRecords = _consumerRecordsProcessor.process(rawRecords);
+
+      xRecords = toXRecords(rawRecords);
+      xRecords = _consumerRecordsProcessor.process(xRecords);
```
smccauliff: That means garbage hangs around for longer.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
```diff
@@ -17,15 +17,16 @@
  */
 public class LargeMessage {
   private final Map<Integer, ByteBuffer> _segments;
-  private final int _messageSize;
+  private final int _originalValueSize;
```
smccauliff: As someone that had not read very much of the internals of Kafka I found the term "message" very confusing because (as far a I can tell) there is no such class.  Even the concept of message is fuzzy.  Does it mean just the value?  Or the key and the value?  What about the timestamp, headers and anything else that gets added to the ConsumerRecord or ProducerRecord class in the future?  So does message mean record?  We could serialize the entire record and then chunk that into smaller segments.  What is true is that large message support actually means sending large values that are part of records.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
```diff
@@ -72,43 +73,45 @@ public long startingOffset() {
...
+    if (_bufferedBytes != _originalValueSize) {
       throw new InvalidSegmentException("Buffered bytes in the message should equal to message size."
-          + " Buffered bytes = " + _bufferedBytes + "message size = " + _messageSize);
+          + " Buffered bytes = " + _bufferedBytes + "message size = " + _originalValueSize);
     }
-    byte[] serializedMessage = new byte[_messageSize];
+    byte[] serializedMessage = new byte[_originalValueSize];
     int segmentStart = 0;
     for (int i = 0; i < _numberOfSegments; i++) {
       ByteBuffer payload = _segments.get(i);
```
smccauliff: I would like to remove the word "message" from as many things as possible unless it really has a well defined meaning in Kafka.  So _serializedValue or _payload works for me.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
smccauliff: And what's even more confusing (to me) is that occasionally the word "event" used.

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitter.java
```diff
@@ -4,117 +4,25 @@
...
-
-import java.util.List;
-import java.util.UUID;
+import com.linkedin.kafka.clients.producer.ExtensibleProducerRecord;
+import java.util.Collection;
 
 /**
  * Message splitter for large messages
+ *
+ * TODO: just make this an interceptor interface?
```
smccauliff: I'll remove this comment.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -195,53 +201,84 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
-            _messageSplitter.split(topic, partition, timestamp, messageId, serializedKey, serializedValue);
-        Callback largeMessageCallback = new LargeMessageCallback(segmentRecords.size(), errorLoggingCallback);
-        for (ProducerRecord<byte[], byte[]> segmentRecord : segmentRecords) {
-          future = _producer.send(segmentRecord, largeMessageCallback);
-        }
+
+      Future<RecordMetadata> future = null;
+      int sizeInBytes = (serializedKey == null ? 0 : serializedKey.length) + (serializedValue == null ? 0 : serializedValue.length);
+      boolean useLargeMessageProcessing = _largeMessageEnabled && serializedValue != null && serializedValue.length > _maxMessageSegmentSize;
+      if ((!(producerRecord instanceof ExtensibleProducerRecord) && !useLargeMessageProcessing) ||
```
smccauliff: I agree this is confusing.  When I get into work I will send you the truth table for this.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
smccauliff: OK

src/test/java/com/linkedin/kafka/clients/producer/DocumetationTest.java
smccauliff: There are other tests that test functionality.  The goal of this testto have a simple working example that can be included in the documentation.

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
smccauliff: We spent time making the magic number can be decoded as a valid UTF-8 string.  This tests that it is not a valid UTF-8 string.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
```diff
@@ -72,43 +73,45 @@ public long startingOffset() {
...
       int payloadSize = payload.remaining();
       payload.get(serializedMessage, segmentStart, payloadSize);
       segmentStart += payloadSize;
     }
-    assert (segmentStart == _messageSize);
+    if (segmentStart !=  _originalValueSize) {
+      throw new IllegalStateException("segmentStart " + segmentStart + " != originalValueSize " + _originalValueSize);
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
```diff
@@ -16,54 +17,170 @@
...
    */
-  public byte[] payloadArray() {
-    if (payload.arrayOffset() == 0 && payload.limit() == payload.array().length) {
-      return payload.array();
-    } else {
-      return Arrays.copyOfRange(payload.array(), payload.arrayOffset(), payload.arrayOffset() + payload.limit());
+  public LargeMessageSegment(byte[] headerPayload, ByteBuffer segment) {
+    ByteBuffer headerReader = ByteBuffer.wrap(headerPayload);
+    byte headerVersion = headerReader.get();
+    if (headerVersion != CURRENT_VERSION) {
```
smccauliff: In open source there will not be other versions.  Internally we can have our own implementation that recognizes the old version and either produces an empty map or a single header which describes the large message segment.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
smccauliff: |Record | Value | Headers | Action                            |
|-------|-------|---------|-----------------------------------|
| p     |   l   |   y     | Can't have headers with p         |
| p     |   l   |   n     | segment large value (convert to x)|
| p     |   s   |   y     | Can't have headers with p         |
| p     |   s   |   n     | pass through without modification |
| p     |   u   |   y     | Can't have headers with p         |
| p     |   u   |   n     | pass through without modification |
| x     |   l   |   y     | segment large value               |
| x     |   l   |   n     | segment large value               |
| x     |   s   |   y     | segment large value or header only|
| x     |   s   |   n     | pass throught without modification|
| x     |   u   |   y     | segment large value or header only|
| x     |   u   |   n     | pass through without modification |

smccauliff: OK.  Looks like that will work.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
smccauliff: I'm not sure about this either.  That's fine we can remove it, but we should document the behavior when null keys are used.

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
becketqin: I was actually suggesting we put the word "Test" in the name so it is more consistent.

smccauliff: Sounds good.

src/test/java/com/linkedin/kafka/clients/producer/DocumetationTest.java
becketqin: So maybe DocumentationExampleTest? It is not testing the documentation itself, because there is actually documentation test in Kafka which tests the auto generated documentation for configurations.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
becketqin: Sure, let's document it then.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
becketqin: I see. Let's use value then.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
becketqin: Hmm, a local variable in this case is just a pointer on stack and literally does not really change anything, right?

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
```diff
@@ -29,8 +32,11 @@
   public static final String AUDITOR_CLASS_CONFIG = "auditor.class";
   public static final String KEY_SERIALIZER_CLASS_CONFIG = ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG;
   public static final String VALUE_SERIALIZER_CLASS_CONFIG = ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG;
-  public static final String SEGMENT_SERIALIZER_CLASS_CONFIG = "segment.serializer";
   public static final String CURRENT_PRODUCER = "current.producer";
+  public static final String UUID_FACTORY_CLASS_CONFIG = "uuid.factory.class";
+  public static final String HEADER_PARSER_CONFIG = LiKafkaConsumerConfig.HEADER_PARSER_CONFIG;
+
+  public static final String HEADER_PARSER_DOC = LiKafkaConsumerConfig.HEADER_PARSER_DOC;
```
smccauliff: Having the serializer code with the deserializer code makes it easier to compare them so you know the serializer and deserializer are working together correctly.  I'm sure why Kafka has them separated other than to make misconfiguration more likely.

becketqin: There is probably not substantial difference between the two. But there might be some slight benefits to separate the interface.

First of all, having separate interface does not prevent people from putting the serde code in the same class if they want to. i.e. implement both interface.

The benefit of having separate interface is that users are not forced to implement both serializer and deserializer if they only want to have one of them, but not both. Imagine we are doing migration and currently we have old wire protocol and now we are going to switch to the new wire protocol with new producer. Likely we only need to let the consumer to deserialize with both new and old wire protocol, so only a customized deserializer is needed, but if we have only one serde interface, users will be forced to implement the serializer as well, which is not ideal.

The configuration on producer and consumer are likely in different processes, so they will have to be configured separately even if the serde code is in the same class. So having serde code in the same class does not really avoid misconfiguration. It is probably more important to have corresponding tests to make sure everything works fine end to end instead of relying on the colocation of the code.

src/main/java/com/linkedin/kafka/clients/consumer/ExtensibleConsumerRecord.java
smccauliff: I went through the javadocs for ExtensibleConsumerRecord, ExtensibleProducerRecord and added documentation for public methods and edited some of the existing documentation.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
smccauliff: New local variable is used.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -257,10 +248,78 @@ public void unsubscribe() {
...
+    for (ExtensibleConsumerRecord<byte[], byte[]> xRecord : xRecords) {
+      ExtensibleConsumerRecord<K, V> userRecord = deserialize(xRecord);
+      if (_auditor != null) {
+        long totalBytes = userRecord.serializedKeySize() + userRecord.serializedValueSize();
+        _auditor.record(userRecord.topic(), userRecord.key(), userRecord.value(), userRecord.timestamp(), 1L, totalBytes,
+            AuditType.SUCCESS);
+      }
+
+      TopicPartition topicPartition = new TopicPartition(userRecord.topic(), userRecord.partition());
+      List<ConsumerRecord<K, V>> listForTopicPartition = consumerRecordsMap.get(topicPartition);
```
smccauliff: Didn't know about that Java 8 feature.  Cool.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
smccauliff: end offset is not tracked anywhere in this class or method.  I've added everything else.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
smccauliff: The version number her is to distinguish future versions of the headers.  Older wire formats are detected by magic number mismatch.

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitter.java
smccauliff: Removed

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitter.java
```diff
@@ -4,117 +4,25 @@
...
+   * then this may return a singleton collection with just previousRecord.
    */
-  List<ProducerRecord<byte[], byte[]>> split(String topic,
-                                             Integer partition,
-                                             Long timestamp,
-                                             UUID messageId,
-                                             byte[] key,
-                                             byte[] serializedRecord,
-                                             int maxSegmentSize);
+  Collection<ExtensibleProducerRecord<byte[], byte[]>> split(ExtensibleProducerRecord<byte[], byte[]> previousRecord);
```
smccauliff: Done

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
smccauliff: This change is made but it doesn't seem to save any lines of code.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
smccauliff: This is actually the size without headers so I'm going to keep this.  I've removed all the code that tracks the header size in the consumer, but forgot to remove that same code in the producer so I will now remove that code as well.

src/test/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerIntegrationTest.java
smccauliff: Removed.

src/test/java/com/linkedin/kafka/clients/producer/LiKafkaProducerIntegrationTest.java
smccauliff: OK

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
smccauliff: Fine.

src/test/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializerDeserializerTest.java
smccauliff: Changed the name of this test.

smccauliff:
Closing to make a new PR with String keys.

PR #21, opened 2017-02-14
minimize compile dependencies to kafka-clients only

radai-rosenblatt:
no more exporting scala onto innocent users.
TBD: rewrite test harnesses to get rid of scala entirely


benmccann:
Should we also add a couple lines to export a test jar? I think we have other code that needs the test harnesses.  http://mrhaki.blogspot.com/2010/11/gradle-goodness-create-jar-artifact.html

Thanks for doing this! This will be an enormous help!!

radai-rosenblatt:
i seriously dislike the current test harness code.
I'd like to convert it to java and move it to be its own module (test-support? test-harness?). 
that way there'd be no need to produce a test artifact (as a test artifact would include our actual tests, not just the harness)

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,10 +1,6 @@
...
 See License in the project root for license information.
-#
-
-#Fri Apr 22 16:57:09 PDT 2016
+#Mon Feb 13 17:24:15 PST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.2-bin.zip
```
benmccann: do you really have to change this line?

radai-rosenblatt: the switch from *-bin to *-all is meant to enable auto-complete in IDEA for gradle build files. the artifact is larger, but its a one-time setup cost and I think the end result is worth it

benmccann: Sounds good then. Thanks for the clarification

PR #25, opened 2017-02-14
Build test jar

benmccann:
Huge thanks to Radai for https://github.com/linkedin/li-apache-kafka-clients/pull/21, which keeps the test harnesses out of the main artifact.

We'll need the test harnesses to be published to an artifact as there are currently tests in other libraries depending on the test harnesses

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,6 @@ sourceSets {
 }
 
 dependencies {
-
     compile 'org.apache.kafka:kafka-clients:0.10.1.1'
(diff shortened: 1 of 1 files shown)


smccauliff:
This compiles and runs all the unit tests after I apply the patch to my working copy.  Thanks for the patch.

benmccann:
Np. How do we get this imported internally? It'd be a really huge help to me to get Radai's fix for my project

PR #26, opened 2017-02-15
build piping for publishing artifacts

radai-rosenblatt:
Artifactory creds taken from env vars.
Also bumped up gradle and checkstyle versions and fixed up javadoc errors.


smccauliff:
@radai-rosenblatt This looks good except there are changes to some of the java files that are not build files. In the immediate future can we stop making changes to the files that don't need to change because I grow weary of constantly having to fix conflicts with that PR.

PR #27, opened 2017-02-16
Distrbute jars to artifactory/jcenter.

smccauliff:
This provides all the mechanisms we need to be able to publish jars into jcenter and linkedin.


buildSrc/src/main/groovy/com/linkedin/gradle/build/DistributeTask.groovy~
```diff
@@ -0,0 +1,40 @@
+package com.linkedin.gradle.build
```
becketqin: It seems this file is not needed. Can we remove this?

PR #29, opened 2017-02-17
drop scala daemon, disable kafka test broker log compaction to save memory

radai-rosenblatt:
for great justice

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,4 @@
 version=0.0.5-SNAPSHOT
+org.gradle.daemon=false
+org.gradle.parallel=false
+org.gradle.jvmargs=-Xms512m -Xmx512m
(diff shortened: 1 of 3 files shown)


radai-rosenblatt:
@smccauliff @becketqin merge all zig ?

smccauliff:
Do we know if (or by how much) this slows down building on local workstations?  I still had to tune the heap size to get this to build on circleci so I don't know how much use this will be.

gradle.properties
```diff
@@ -1 +1,3 @@
-version=0.0.1-SNAPSHOT
\ No newline at end of file
+version=0.0.1-SNAPSHOT
```
becketqin: Should this be updated? We have 0.0.4 published officially.

PR #31, opened 2017-02-17
Fix the issue that null value messages are not handled correctly.

becketqin:


src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -311,10 +314,12 @@ public void close() {
     K key = _keyDeserializer.deserialize(tp.topic(), consumerRecord.key());
     byte[] valueBytes = parseAndMaybeTrackRecord(tp, consumerRecord.offset(), consumerRecord.value());
     V value = _valueDeserializer.deserialize(tp.topic(), valueBytes);
-    if (value != null) {
+    if (valueBytes != INCOMPLETE_RESULT) {
```
smccauliff: V value probably needs to get initialized inside of the check for if (valueBytes != INCOMPLETE_RESULT) since we don't know what the deserializer will do with it; it may throw an exception.

smccauliff:
I just have that one comment.  Otherwise TGTM!

src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
becketqin: The handling of the value are user code. We should use whatever value returned by the deserializer. It would be a little weird to have any value other than that. Also we do not call any method on the value so it is safe.

PR #33, opened 2017-02-18
Document how to add the artifact repository containing the li-apache-kafka-jars.

smccauliff:
Cut and paste directions for adding the jars as a dependency to your gradle project.


radai-rosenblatt:
the instructions will work, but I feel we should just complete the piping and get this into central.

smccauliff:
The artifacts got pushed to jcenter; that seems like an acceptable repository for an open source project.  No?

PR #34, opened 2017-02-22
Header API.

smccauliff:


src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializer.java
```diff
@@ -0,0 +1,118 @@
...
+   * @return VERSION_AND_FLAGS_SIZE if headers is null else the number of bytes needed to represent the header key and value, but
+   * without the magic number.
+   */
+  @Override
+  public int serializedHeaderSize(Map<String, byte[]> headers) {
+    if (headers == null) {
+      return DefaultHeaderSerializer.VERSION_AND_FLAGS_SIZE;
+    }
+    // size of all the keys and the value length fields
+    int size = headers.size() * 5 + VERSION_AND_FLAGS_SIZE + ALL_HEADER_SIZE_FIELD_SIZE + _headerMagic.length;
```
becketqin: Magic value 5.

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerde.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+package com.linkedin.kafka.clients.utils;
+
+import java.nio.ByteBuffer;
+
+
+/**
+ * Utilities to implement header serializer and deserializer.
+ */
+public interface DefaultHeaderSerde {
```
becketqin: Can we either move some (probably all) of the things in this class to HeaderUtils and rename this class to something like HeaderConstants?

src/main/java/com/linkedin/kafka/clients/producer/ExtensibleProducerRecord.java
```diff
@@ -0,0 +1,149 @@
...
+ * consumer must be using {@link com.linkedin.kafka.clients.consumer.LiKafkaConsumer}.  If headers are present and the
+ * consumer is not using
+ * {@link com.linkedin.kafka.clients.consumer.LiKafkaConsumer} then the consumer records will not be deserializable.
+ * True null values can not be sent in the presence of any headers.
+ * </p>
+ * <p>{@link HeaderUtils} contains suggestions for how to partition the header key into
+ * intervals.  Header keys must be non-null, non-zero length strings that can be encoded into a utf-8 string.  Headers
+ * live in a separate space from the underlying Kafka protocol.
+ * </p>
+ * <p> A header can be added to a producer record by calling {@link #header(String, byte[])}.  This implies that the
```
smccauliff: In the consumer record:
We want headers to be mutable because we want to have plugins at some point that support modification of headers.  Right now I need this in order to remove the large message header otherwise we could just leave the header there.  Since the large message support is not transparent to the end user I think it's fine to just leave the large message header.

In the producer record (which this comment specifically concerns):
We hide the data structure used to store the headers from the user so either we expose that we are using a Map or else we have some kind of HeaderBuilder that builds a Collection<Header> that is then passed to the producer constructor.

Since we have byte[] values, unless we do expensive defensive copying for each record, the record is still mutable.

src/main/java/com/linkedin/kafka/clients/producer/ExtensibleProducerRecord.java
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * A simple predicate that returns true if there are any header key-value mappings on this record.
+   * @return true if this record contains any header key value pairs else false
+   */
+  public boolean hasHeaders() {
+    return headers != null && !headers.isEmpty();
+  }
+
+  /**
+   *
```
smccauliff: Explained.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
```diff
@@ -29,8 +32,11 @@
   public static final String AUDITOR_CLASS_CONFIG = "auditor.class";
   public static final String KEY_SERIALIZER_CLASS_CONFIG = ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG;
   public static final String VALUE_SERIALIZER_CLASS_CONFIG = ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG;
-  public static final String SEGMENT_SERIALIZER_CLASS_CONFIG = "segment.serializer";
   public static final String CURRENT_PRODUCER = "current.producer";
+  public static final String UUID_FACTORY_CLASS_CONFIG = "uuid.factory.class";
+  public static final String HEADER_SERIALIZER_CONFIG = "header.serializer.class";
+
+  public static final String HEADER_PARSER_DOC = LiKafkaConsumerConfig.HEADER_PARSER_DOC;
```
smccauliff: Thanks for catching this.  Something I missed during refactoring.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
       V value = producerRecord.value();
       Object auditToken = _auditor.auditToken(key, value);
```
smccauliff: Let's discuss changes to the audit API with Radai when he gets back.  This is a change I would push out to some future PR.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
         _auditor.record(auditToken, topic, timestamp, 1L, 0L, AuditType.FAILURE);
         throw new KafkaException(t);
       }
-      int sizeInBytes = (serializedKey == null ? 0 : serializedKey.length)
-          + (serializedValue == null ? 0 : serializedValue.length);
-      // Audit the attempt.
+
+      int sizeInBytes = (serializedKey == null ? 0 : serializedKey.length) + (serializedValue == null ? 0 : serializedValue.length);
+      boolean useLargeMessageProcessing = _largeMessageEnabled && serializedValue != null && serializedValue.length > _maxMessageSegmentSize;
+      if (!(producerRecord instanceof ExtensibleProducerRecord) && serializedValue == null) {
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
+          sizeInBytes, auditToken);
+      }
+
+      if (producerRecord instanceof ExtensibleProducerRecord && !((ExtensibleProducerRecord) producerRecord).hasHeaders() &&
+        serializedValue == null) {
+        return sendNullValue(producerRecord, callback, topic, key, value, timestamp, serializedValue, serializedKey,
+          sizeInBytes, auditToken);
+      }
+
+      ExtensibleProducerRecord<byte[], byte[]> xRecord =
```
smccauliff: A boolean to allow for mutability?

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
-                      producerRecord.timestamp(), 1L, 0L, AuditType.FAILURE);
+          producerRecord.timestamp(), 1L, 0L, AuditType.FAILURE);
       throw new KafkaException(t);
     } finally {
       _numThreadsInSend.decrementAndGet();
     }
   }
 
+  private Future<RecordMetadata> sendNullValue(ProducerRecord<K, V> producerRecord, Callback callback, String topic,
+    K key, V value, Long timestamp, byte[] serializedValue, byte[] serializedKey, int sizeInBytes, Object auditToken) {
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
+    future = _producer.send(headerlessByteRecord, errorLoggingCallback);
+    return future;
+  }
+
+  /**
+   * This is public for testing.
+   * This assumes that the value is not a large value that would be segmented by large message support.
+   */
+  public static ProducerRecord<byte[], byte[]> serializeWithHeaders(ExtensibleProducerRecord<byte[], byte[]> xRecord,
+    HeaderSerializer headerParser) {
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -325,9 +374,9 @@ public ErrorLoggingCallback(UUID messageId,
     @Override
     public void onCompletion(RecordMetadata recordMetadata, Exception e) {
       if (e != null) {
-        LOG.error(String.format("Unable to send event %s with message id %s to kafka topic %s",
+        LOG.error(String.format("Unable to send record %sto kafka topic %s",
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
```diff
@@ -304,11 +284,12 @@ public void clear(TopicPartition tp) {
     _partitionConsumerHighWatermark.remove(tp);
   }
 
+
   public void close() {
     _messageAssembler.close();
-    _auditor.close();
   }
 
+  /*
```
smccauliff: oops.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
+    HeaderSerializer headerParser) {
+
+    if (!xRecord.hasHeaders()) {
+      return new ProducerRecord<byte[], byte[]>(xRecord.topic(), xRecord.partition(), xRecord.timestamp(), xRecord.key(), xRecord.value());
+    }
+
+    int headersSize = headerParser.serializedHeaderSize(xRecord.headers());
+    int serializedValueSize = (xRecord.value() == null ? 0 : xRecord.value().length) + headersSize;
+
+    ByteBuffer valueWithHeaders = ByteBuffer.allocate(serializedValueSize);
```
smccauliff: Let's discuss this for a followup patch.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
+    headerParser.serializeHeader(valueWithHeaders, xRecord.headers(), xRecord.value() == null);
```
smccauliff: This method can't be moved to HeaderUtils because it is dependent on the header serializer used.  I'm not sure how value can be a ByteBuffer since to avoid copying the ByteBuffer the header is seriailzed into needs to be serializedValue.length + serializedHeader.length.

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
```diff
@@ -333,34 +314,60 @@ public void close() {
           valueBytes == null ? 0 : valueBytes.length,
           _keyDeserializer.deserialize(consumerRecord.topic(), consumerRecord.key()),
           value);
+*/
+
+  /**
+   * When we encounter an incomplete large message this filters (i.e. it returns null) otherwise this tracks records and
+   * returns completed large messages and non-large messages.
+   * @param srcRecord this is the record that was polled from the underlying Kafka consumer.
+   * @return this may return null
```
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
```diff
@@ -333,34 +314,60 @@ public void close() {
...
+    byte[] key = assembledResult.isOriginalKeyNull() ? null : srcRecord.key();
+    int serializedValueSize = assembledResult.messageBytes().length;
+
+    ExtensibleConsumerRecord<byte[], byte[]> largeMessageRecord =
+      new ExtensibleConsumerRecord<>(srcRecord.topic(), srcRecord.partition(), srcRecord.offset(),
+        srcRecord.timestamp(), srcRecord.timestampType(),
+        srcRecord.checksum(),
+        serializedKeySize, serializedValueSize,
+        key, assembledResult.messageBytes());
+    //TODO: checksums recomputed?
```
smccauliff: It's not clear to me why someone would want the checksum since they would need to recompute the original bytes that were sent in order to verify the checksum and presumably the underlying code has already done that.

smccauliff: I have removed this comment.

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerde.java
smccauliff: All these constants really have to do with how the header is encoded and not constants that deal with all possible encodings.  For example, you could encode the headers as UTF-16 which is Java's native encoding which means more overhead for ascii characters, but less copying on the consumer side.  Magic numbers are very specific to our encoding.   I don't see anything that belongs in HeaderUtils.

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializer.java
smccauliff: Ok

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
```diff
@@ -16,54 +17,170 @@
...
    */
-  public byte[] payloadArray() {
-    if (payload.arrayOffset() == 0 && payload.limit() == payload.array().length) {
-      return payload.array();
-    } else {
-      return Arrays.copyOfRange(payload.array(), payload.arrayOffset(), payload.arrayOffset() + payload.limit());
+  public LargeMessageSegment(byte[] headerPayload, ByteBuffer segment) {
+    ByteBuffer headerReader = ByteBuffer.wrap(headerPayload);
+    byte headerVersion = headerReader.get();
+    if (headerVersion != CURRENT_VERSION) {
```
smccauliff: There is no previous version of this so we are ok with not accepting a version 0 (although I realize we kind of have another version of this with a different magic number, but that should be handled separately).  By accepting version >= 1 then all future versions must be wire format compatible with version 1.  That seems like a very big restriction.  If for some reason future versions are wire compatible with version 1 then we can branch the code containing version 1 and allow it to deserialize the newer version.  This does mean updating all the consumers that are running the code, however, the new code won't be doing anything algorithmically different from the previous consumer code.

src/main/java/com/linkedin/kafka/clients/producer/ExtensibleProducerRecord.java
```diff
@@ -0,0 +1,149 @@
...
+   */
+  public void header(String headerKey, byte[] headerValue) {
+    HeaderUtils.validateHeaderKey(headerKey);
+
+    if (headerValue == null) {
+      throw new IllegalArgumentException("Header value must not be null.");
+    }
+
+    if (headers == null) {
+      headers = new TreeMap<>();
```
smccauliff: When we had integer keys TreeMap had slightly better performance.  For String keys HashMap has slightly better performance.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageBufferPool.java
```diff
@@ -163,35 +163,16 @@ private LargeMessage evictEldestMessage() {
...
     LargeMessage message = _incompleteMessageMap.get(messageId);
     if (message == null) {
-      message = new LargeMessage(tp, messageId, offset, messageSizeInBytes, numberOfSegments);
+      message = new LargeMessage(tp, messageId, offset, segment.originalValueSize(), segment.numberOfSegments());
       _incompleteMessageMap.put(messageId, message);
       LOG.trace("Incomplete message {} is created.", messageId);
     }
+
     if (message.startingOffset() > offset) {
       throw new InvalidSegmentException("Out of order segment offsets detected.");
```
smccauliff: I've put the offset into various exception messages.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -196,61 +198,115 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
-        future = _producer.send(wrappedRecord.get(0), errorLoggingCallback);
+        new ErrorLoggingCallback<>(auditToken, topic, timestamp, sizeInBytes, _auditor, callback);
+
+      if (useLargeMessageProcessing) {
+        errorLoggingCallback = new LargeMessageCallback(xRecords.size(), errorLoggingCallback);
+      }
+      Future<RecordMetadata> future = null;
+      for (ExtensibleProducerRecord<byte[], byte[]> segmentRecord : xRecords) {
+        ProducerRecord<byte[], byte[]> segmentProducerRecord = serializeWithHeaders(segmentRecord, _headerSerializer);
+        future = _producer.send(segmentProducerRecord, errorLoggingCallback);
```
smccauliff: Let's do a separate PR.

smccauliff:
I've taken care of everything I think should be taken care of or else I believe it should be done in a separate PR.

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
```diff
@@ -304,63 +284,66 @@ public void clear(TopicPartition tp) {
...
+      //Not a complete, large message
+      _deliveredMessageOffsetTracker.addNonMessageOffset(topicPartition, srcRecord.offset());
+      return null;
     }
+
+    // Completed, large message value
+    _deliveredMessageOffsetTracker.track(topicPartition, srcRecord.offset(), safeOffset, assembledResult.messageStartingOffset(),
+        true);
+
+    int serializedKeySize = assembledResult.isOriginalKeyNull() ? 0 : srcRecord.key().length;
```
radai-rosenblatt: micro - optimization: checking the same bool cond. twice.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -504,6 +565,32 @@ public void close() {
     }
     _kafkaConsumer.close();
```
radai-rosenblatt: this would benefit greatly from a util ,ethod that would ensure all close() methods are called even in the presence of exceptions

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -504,6 +565,32 @@ public void close() {
...
+    }
+    try {
+      Method timedCloseMethod = _kafkaConsumer.getClass().getDeclaredMethod("close", Long.TYPE, TimeUnit.class);
+      timedCloseMethod.invoke(_kafkaConsumer, timeout, unit);
+    } catch (NoSuchMethodException | IllegalAccessException cantCallTimedCloseMethod) {
+      _kafkaConsumer.close();
+    } catch (InvocationTargetException ite) {
+      throw new KafkaException(ite.getTargetException());
+    }
+    _consumerRecordsProcessor.close();
```
radai-rosenblatt: same util.closeAll()

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessage.java
```diff
@@ -39,12 +40,12 @@ public synchronized long bufferedSizeInBytes() {
   }
 
   public synchronized SegmentAddResult addSegment(LargeMessageSegment segment, long offset) {
```
radai-rosenblatt: +1 on starting offset. i'd leave the offset in case we ever want to support out-of-order arrival of segments

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
radai-rosenblatt: i would replace "invalid" in the exception text with "unsupported"

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
```diff
@@ -16,54 +17,170 @@
...
+    this._sequenceNumber = headerReader.getInt();
+    if (_sequenceNumber < 0) {
+      throw new InvalidSegmentException("Sequence number must be non-negative.");
+    }
+    this._numberOfSegments = headerReader.getInt();
+    if (_numberOfSegments < 0) {
+      throw new InvalidSegmentException("Number of segments must be non-negative.");
+    }
+    if (_numberOfSegments <= _sequenceNumber) {
+      throw new InvalidSegmentException("Number of segments must be greater than sequence number.");
```
radai-rosenblatt: could print the offending value(s)

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
```diff
@@ -16,54 +17,170 @@
...
     }
+    this._originalValueSize = headerReader.getInt();
+    if (_originalValueSize < 0) {
+      throw new InvalidSegmentException("Original value size must be non-negative.");
```
radai-rosenblatt: could print the offending value(s)

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
```diff
@@ -4,93 +4,72 @@
...
 import java.util.List;
 import java.util.UUID;
 
 /**
- * This class is used by {@link com.linkedin.kafka.clients.producer.LiKafkaProducerImpl} to split to split
+ * This class is used by {@link com.linkedin.kafka.clients.producer.LiKafkaProducerImpl} to split
  * serialized records.
  */
-public class MessageSplitterImpl implements MessageSplitter {
+public class MessageSplitterImpl<K, V> implements MessageSplitter<K, V> {
```
radai-rosenblatt: why is this class generic? splitting happens post-serialization ... ?

src/main/java/com/linkedin/kafka/clients/producer/ExtensibleProducerRecord.java
radai-rosenblatt: headers should be mutable :-)

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerde.java
```diff
@@ -0,0 +1,79 @@
...
+    bbuf.putLong(DEFAULT_HEADER_MAGIC);
+    return bbuf.array();
+  }
+
+  /**
+   * The length of a utf8 encoded string.
+   * @param s non-null
+   * @return a non-negative integer. The number of bytes in the utf-8 encoded form of the string
+   */
+  static int utf8StringLength(String s) {
```
radai-rosenblatt: seems to me code using this could be refactored to actually serialize the string and look at the resulting length? do we ever need the size without needing the serialized string?

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializer.java
```diff
@@ -0,0 +1,118 @@
...
+    int originalPosition = dest.position();
+    dest.put(_headerMagic);
+    byte versionAndFlags = (byte) (VERSION_1 | (userValueIsNull ? USER_VALUE_IS_NULL_FLAG : 0));
+    dest.put(versionAndFlags);
+    dest.putInt(0); //updated later
+    if (headers == null) {
+      return;
+    }
+
+    CharsetEncoder utf8Encoder = StandardCharsets.UTF_8.newEncoder();
```
radai-rosenblatt: thread local FTW?

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerde.java
smccauliff: I'm trying to avoid serializing the string twice: once to get the size and the second time to serialize it into byte buffer.

src/main/java/com/linkedin/kafka/clients/largemessage/LargeMessageSegment.java
smccauliff: OK

build.gradle
```diff
@@ -39,14 +40,17 @@ sourceSets {
...
-//project dependencies
-
+// dependencies
 repositories {
     mavenCentral()
 }
 
 dependencies {
-    compile 'org.apache.kafka:kafka-clients:0.10.1.1'
+    compile 'org.apache.kafka:kafka_2.10:0.10.2.0:test'
```
smccauliff: Will try it.

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
```diff
@@ -58,24 +43,19 @@ public ConsumerRecordsProcessor(MessageAssembler messageAssembler,
...
-      // Only put record into map if it is not null
-      if (handledRecord != null) {
-        List<ConsumerRecord<K, V>> list = filteredRecords.get(tp);
-        if (list == null) {
-          list = new ArrayList<>();
-          filteredRecords.put(tp, list);
-        }
-        list.add(handledRecord);
+  public Collection<ExtensibleConsumerRecord<byte[], byte[]>> process(Collection<ExtensibleConsumerRecord<byte[], byte[]>> consumerRecords) {
+    List<ExtensibleConsumerRecord<byte[], byte[]>> list = new ArrayList<>();
```
smccauliff: How about initializing to consumerRecords.size()?

src/main/java/com/linkedin/kafka/clients/consumer/ConsumerRecordsProcessor.java
smccauliff: Avoiding this double checking will actually result in more lines of code.  Also both of these are conditional assignment statements which should get compiled to CMOV instructions which are easily pipelined since the CPU does not need to do any branch prediction on CMOV.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
smccauliff: OK.  Look at my solution to this problem in the next commit.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
smccauliff: OK

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
smccauliff: I forgot to remove all the user key, value types.

src/main/java/com/linkedin/kafka/clients/utils/DefaultHeaderSerializer.java
smccauliff: OK and on the decode side as well.

src/test/java/com/linkedin/kafka/clients/producer/DocumentationTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+
+/**
+ * This is a simple code example we use in the documentation.
+ */
+public class DocumentationTest extends AbstractKafkaClientsIntegrationTestHarness {
```
smccauliff: Sure.

src/main/java/com/linkedin/kafka/clients/utils/HeaderDeserializer.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+package com.linkedin.kafka.clients.utils;
```
smccauliff: Moved to headers package.

smccauliff:
Mege?

PR #37, opened 2017-03-23
Fix seekToEnd

becketqin:


src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -229,12 +229,12 @@ public void unsubscribe() {
       } catch (OffsetOutOfRangeException | NoOffsetForPartitionException oe) {
         switch (_offsetResetStrategy) {
           case EARLIEST:
-            _kafkaConsumer.seekToBeginning(oe.partitions());
             oe.partitions().forEach(_consumerRecordsProcessor::clear);
```
smccauliff: If EARLIEST or LATEST is in the interval of offsets tracked by the consumer records processor then should this clear the state of the consumer records processor?

becketqin: Not sure I understand what you meant. This catch clause is to handle the case where broker returns an OffsetOutOfRangeException and the consumer has a offset reset strategy configured. In this case, reset the offset to either earliest and latest as if we haven't consumed any message.

src/test/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerIntegrationTest.java
```diff
@@ -265,36 +265,43 @@ public void testCommit() {
...
-        records = consumer.poll(100);
-      }
-      assertEquals(records.iterator().next().offset(), 2L, "Should see message offset 2 again.");
-    } finally {
-      consumer.close();
+      ConsumerRecords<String, String> records = consumer.poll(1000L);
+      assertEquals(records.iterator().next().offset(), 2L, "Should see message offset 2.");
+
+      consumer.seekToEnd(Collections.singleton(tp));
+      consumer.poll(10L);
```
smccauliff: You might want to make this 100L or something so it's not confused with the offset that is being looked at next.  Otherwise LGMT.

becketqin: If we make it 100L it will actually block for 100L here because we just seek to the end. I think it is OK as it should be a well established concept that consumer.poll() takes a timeout instead offset.

becketqin:
@smccauliff Thanks for the review. Merged to the trunk.

PR #38, opened 2017-03-30
Fix partition-level commitAsync API.

jjkoshy:


becketqin:
Thanks for the patch. LGTM.

PR #40, opened 2017-04-26
Change to use SHA1PRNG for UUID generation. Catch exception and close the producer/consumer if LiKafkaProducer/LiKafkaConsumer creation failed

becketqin:


smccauliff:
If there is a UUID factory then users can just implement their own UUID some of which can be generated much faster than this one.

smccauliff:
I must have missed where messageId gets passed into the splitter.  
LGTM!

PR #41, opened 2017-05-10
Auditor#configure will be invoked twice if configured using auditor.class property

apakulov:
Hi LinkedIn,
I found a small issue with Auditor initialization. It will be invoked twice if you're configuring it using **auditor.class** property. In producer's case, it will invoke **configure** without current_producer assigned and the second time with it.

Testing code is a little bit verbose since there is no mocking library and didn't want to bring it with my little change.

Thank you!


becketqin:
@apakulov Thanks for reporting the issue. Yes, that is a problem. Would it be simpler to let LiKafkaProducerConfig to override the `getConfiguredInstance` method and always pass in the producer to the configurables?

I am also thinking that it might be clearer to just not pass in the underlying producer. Do you plan to use the shared producer?

apakulov:
@becketqin yes I was planning on using underlying producer in Auditor instance.

As for `getConfiguredInstance` method, it will fix the issue when there is no producer passed as config parameter, but `configured` will be invoked the second time explicitly on auditor's instance.
I think we still want to keep two ways of initializing auditor separately.

Thanks and let me know if you want me to fix this code differently.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -114,7 +114,7 @@ private LiKafkaConsumerImpl(LiKafkaConsumerConfig configs,
     _lastAutoCommitMs = System.currentTimeMillis();
     // We need to set the auto commit to false in KafkaConsumer because it is not large message aware.
     ByteArrayDeserializer byteArrayDeserializer = new ByteArrayDeserializer();
-    _kafkaConsumer = new KafkaConsumer<>(configs.configForVanillaConsumer(),
+    _kafkaConsumer = new KafkaConsumer<>(configs.configForVanillaConsumer().originals(),
```
becketqin: In LiKafkaConsumerConfig we wrapped the map into a `LiKafkaConsumerConfig`, and here we unwrap it. Is there any special consideration?

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -173,8 +173,13 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
       _uuidFactory = configs.getConfiguredInstance(LiKafkaProducerConfig.UUID_FACTORY_CLASS_CONFIG, UUIDFactory.class);
       _messageSplitter = new MessageSplitterImpl(_maxMessageSegmentSize, segmentSerializer, _uuidFactory);
       // Instantiate auditor if necessary
-      _auditor = auditor != null ? auditor
-          : configs.getConfiguredInstance(LiKafkaProducerConfig.AUDITOR_CLASS_CONFIG, Auditor.class);
+      LiKafkaProducerConfig auditorConfigs = configs.configsWithCurrentProducer(_producer);
```
becketqin: This will potentially print the producer configurations twice in the log, which could be a little annoying. I think it might be cleaner to just override the `getConfiguredInstance()` method.

becketqin:
@apakulov Are you still working on this patch? Feel free to let me know if you are busy. Thanks.

apakulov:
@becketqin been sidetracked on Friday, planning to fix it by EOD or tomorrow in morning :-)

becketqin:
@apakulov Cool! Thanks!

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
```diff
@@ -69,4 +70,22 @@
     super(CONFIG, props, false);
   }
 
+  /**
+   * This method won't call {@link Configurable#configure(Map)} for {@link LiKafkaProducerConfig#AUDITOR_CLASS_CONFIG}
+   * during instance creation
+   */
+  @Override
+  public <T> T getConfiguredInstance(String key, Class<T> t) {
+    if (AUDITOR_CLASS_CONFIG.equals(key)) {
```
becketqin: Would this be too specific? I think always call `Configurable.configure()` is the right behavior for `getConfiguredInstance()`, The current problem is that we are not passing in the producer when we call `Configurable.configure()`. So I am thinking that maybe we just need to override `getConfiguredInstance` from `AbstractConfig` in LiKafkaProducerConfig` and make sure when `Configurable.configure()` is invoked,the configuration map with the underlying producer is passed in.

src/main/java/com/linkedin/kafka/clients/common/config/LiAbstractConfig.java
```diff
@@ -0,0 +1,35 @@
+package com.linkedin.kafka.clients.common.config;
+
+import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.utils.Utils;
+
+import java.util.Map;
+
+public abstract class LiAbstractConfig extends AbstractConfig {
```
becketqin: Is this class an overkill if we only need to create the configured instance?

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
apakulov: Not sure how you want to do this just by overriding ```getConfiguredInstance```.

Here is the method itself:
```java
public <T> T getConfiguredInstance(String key, Class<T> t) {
  Class<?> c = getClass(key);
  if (c == null)
    return null;
  Object o = Utils.newInstance(c);
  if (!t.isInstance(o))
    throw new KafkaException(c.getName() + " is not an instance of " + t.getName());
  if (o instanceof Configurable)
    ((Configurable) o).configure(originals());
  return t.cast(o);
}
```

Do you want the new code to look like this? Swap ```originals()``` with ```configsWithCurrentProducer()```. This approach won't work because we don't have ```_producerObject_``` defined in LiKafkaProducerConfig object or passed via parameter.

```java
@Override
public <T> T getConfiguredInstance(String key, Class<T> t) {
  Class<?> c = getClass(key);
  if (c == null)
    return null;
  Object o = Utils.newInstance(c);
  if (!t.isInstance(o))
    throw new KafkaException(c.getName() + " is not an instance of " + t.getName());
  if (o instanceof Configurable)
    ((Configurable) o).configure(configsWithCurrentProducer(_producerObject_));
  return t.cast(o);
}
```

There are few ways to fix this:
1. Introduce method that overloads ```getConfiguredInstance``` with extra parameter:
```public <T> T getConfiguredInstance(String key, Class<T> t, Producer<byte[], byte[]> producer)```
2. Pass producer object into LiKafkaProducerConfig using setter after it get initialized in LiKafkaProducerImpl.

I'm leaning towards the first approach. What do you think @becketqin ?

File: src/main/java/com/linkedin/kafka/clients/common/config/LiAbstractConfig.java (hunk 4 above)
Comment (apakulov) [inline src/main/java/com/linkedin/kafka/clients/common/config/LiAbstractConfig.java]: Agree, but I didn't want to duplicate ```getInstance(String key, Class<T> t)``` implementation in ```LiKafkaProducerConfig``` and ```LiKafkaConsumerConfig``` classes. We can remove it once we agree on the approach above.

File: src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java (hunk 3 above)
Comment (becketqin) [inline src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java]: @apakulov Thanks for the quick response. Yes, I think the first one is cleaner. We don't need to to have this in `LiKafkaConsumerConfig`.

File: src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java [hunk 5]
```diff
@@ -136,8 +136,13 @@ private LiKafkaConsumerImpl(LiKafkaConsumerConfig configs,
     DeliveredMessageOffsetTracker messageOffsetTracker = new DeliveredMessageOffsetTracker(maxTrackedMessagesPerPartition);
 
     // Instantiate auditor if needed.
-    Auditor<K, V> auditor = consumerAuditor != null ? consumerAuditor :
-        configs.getConfiguredInstance(LiKafkaConsumerConfig.AUDITOR_CLASS_CONFIG, Auditor.class);
+    Auditor<K, V> auditor;
```
becketqin: The previous assumption we had was that if an auditor is passed in by the users, we will use it as is. Because users has access to the auditor object and can always call `configure()` by themselves. If we call configure again it may cause configure to be called twice. Maybe we can make the documentation clearer on this?

src/test/java/com/linkedin/kafka/clients/auditing/abstractimpl/ConfigureAuditorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Map;
+import java.util.Properties;
+
+import static org.testng.AssertJUnit.assertEquals;
+
+public class ConfigureAuditorTest extends AbstractKafkaClientsIntegrationTestHarness {
+  @BeforeMethod
+  @Override
+  public void setUp() {
```
becketqin: These `setUp` and `tearDown` methods are actually quite expensive. It would be good to see if we can create producer and consumer without bring up the servers.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
apakulov: Yep, it wasn't clear from the doc, unfortunately. But make sense.
I've fixed the documentation.

src/test/java/com/linkedin/kafka/clients/auditing/abstractimpl/ConfigureAuditorTest.java
apakulov: My bad, it only needs proper brokerUrl configured to work. Fixed.

becketqin:
Thanks for the patch. LGTM.

becketqin:
@apakulov Just want to let you know that I just changed the behavior to also call `auditor.configure` for the passed in auditor. Otherwise the user will not have access to the internal producer if the auditor is constructed externally. Sorry for being back and forth on this.

apakulov:
@becketqin thanks for heads up!

PR #42, opened 2017-05-18
If no key is provided for a message, splitter should not set the key to UUID.

becketqin:


becketqin:
@smccauliff Would you have time to take a look? Thanks.

smccauliff:
LGTM! You might want to wait until the CI has run all the tests, but it should probably be ok.

Issue #43, opened 2017-06-14
Need to handle errors during records processing in ConsumerRecordsProcessor.


becketqin:
We need to handle errors in ConsumerRecordsProcessor so that it can deliver the records in the same batch but before the record with error.

PR #44, opened 2017-06-14
Issue#43 Handle exception in the message processing

becketqin:
This patch is to solve #43 .

### DIFF
--- a/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
+++ b/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
@@ -80,10 +80,11 @@ public class LiKafkaConsumerConfig extends AbstractConfig {
       + "the offset to the latest offset</li><li>none: throw exception to the consumer if no previous offset is found "
       + "for the consumer's group</li><li>anything else: throw exception to the consumer.</li></ul>";
 
-  private static final String SKIP_RECORD_ON_EXCEPTION_DOC = "Skip the record with processing error. This is to allow "
-      + "the users who do are willing to ignore the problematic record to continue consuming without worrying about "
-      + "the error handling. By default it is set to false and an exception will be thrown if record processing "
-      + "encounters an error. If user ignore the exception and call poll again, the error message will also be skipped.";
+  private static final String SKIP_RECORD_ON_EXCEPTION_DOC = "Skip the record with RecordProcessingException. "
+      + "This is to allow the users who do are willing to ignore the problematic record to continue consuming without "
+      + "worrying about the error handling. By default it is set to false and an exception will be thrown if record "
+      + "processing encounters an error. If user ignore the exception and call poll again, the error message will also "
+      + "be skipped.";
 
   static {
     CONFIG = new ConfigDef()
--- a/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
+++ b/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
@@ -35,6 +35,7 @@ public class LiKafkaConsumerConfig extends AbstractConfig {
   public static final String ENABLE_AUTO_COMMIT_CONFIG = ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG;
   public static final String AUTO_COMMIT_INTERVAL_MS_CONFIG = ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG;
   public static final String AUTO_OFFSET_RESET_CONFIG = ConsumerConfig.AUTO_OFFSET_RESET_CONFIG;
+  public static final String SKIP_RECORD_ON_EXCEPTION_CONFIG = "skip.record.on.exception";
 
   private static final String MESSAGE_ASSEMBLER_BUFFER_CAPACITY_DOC = "The maximum number of bytes the message assembler " +
       " uses to buffer the incomplete large message segments. The capacity is shared by messages from all the topics. " +
@@ -78,6 +79,11 @@ public class LiKafkaConsumerConfig extends AbstractConfig {
       + "<ul><li>earliest: automatically reset the offset to the earliest offset<li>latest: automatically reset "
       + "the offset to the latest offset</li><li>none: throw exception to the consumer if no previous offset is found "
       + "for the consumer's group</li><li>anything else: throw exception to the consumer.</li></ul>";
+
+  private static final String SKIP_RECORD_ON_EXCEPTION_DOC = "Skip the record with processing error. This is to allow "
+      + "the users who do are willing to ignore the problematic record to continue consuming without worrying about "
+      + "the error handling. By default it is set to false and an exception will be thrown if record processing "
+      + "encounters an error. If user ignore the exception and call poll again, the error message will also be skipped.";
 
   static {
     CONFIG = new ConfigDef()
@@ -138,7 +144,12 @@ public class LiKafkaConsumerConfig extends AbstractConfig {
                 Type.STRING,
                 "none",
                 Importance.MEDIUM,
-                AUTO_OFFSET_RESET_DOC);
+                AUTO_OFFSET_RESET_DOC)
+        .define(SKIP_RECORD_ON_EXCEPTION_CONFIG,
+                Type.BOOLEAN,
+                "false",
+                Importance.LOW,
+                SKIP_RECORD_ON_EXCEPTION_DOC);
 
   }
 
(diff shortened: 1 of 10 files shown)


becketqin:
ping @radai-rosenblatt @smccauliff

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -148,16 +150,20 @@ private LiKafkaConsumerImpl(LiKafkaConsumerConfig configs,
...
-                                                               messageOffsetTracker, auditor);
+                                                               messageOffsetTracker, auditor, skipRecordOnException);
 
     // Instantiate consumer rebalance listener
     _consumerRebalanceListener = new LiKafkaConsumerRebalanceListener<>(_consumerRecordsProcessor,
                                                                         this, _autoCommitEnabled);
 
     // Instantiate offset commit callback.
     _offsetCommitCallback = new LiKafkaOffsetCommitCallback();
+    _lastProcessedResult = null;
```
radai-rosenblatt: not needed ?

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -79,6 +80,11 @@
       + "the offset to the latest offset</li><li>none: throw exception to the consumer if no previous offset is found "
       + "for the consumer's group</li><li>anything else: throw exception to the consumer.</li></ul>";
 
+  private static final String SKIP_RECORD_ON_EXCEPTION_DOC = "Skip the record with processing error. This is to allow "
+      + "the users who do are willing to ignore the problematic record to continue consuming without worrying about "
+      + "the error handling. By default it is set to false and an exception will be thrown if record processing "
+      + "encounters an error. If user ignore the exception and call poll again, the error message will also be skipped.";
```
smccauliff: Might want to be more specific about what kind of Exceptions are caught.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
becketqin: Just a good habit to always initialize variables.

src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -58,22 +62,25 @@ public ConsumerRecordsProcessor(MessageAssembler messageAssembler,
...
     for (ConsumerRecord<byte[], byte[]> record : consumerRecords) {
       TopicPartition tp = new TopicPartition(record.topic(), record.partition());
-      ConsumerRecord<K, V> handledRecord = handleConsumerRecord(record);
-      // Only put record into map if it is not null
-      if (handledRecord != null) {
-        List<ConsumerRecord<K, V>> list = filteredRecords.get(tp);
-        if (list == null) {
-          list = new ArrayList<>();
-          filteredRecords.put(tp, list);
+      if (result.hasError(tp)) {
```
radai-rosenblatt: do we have to try and continue processing records for partitions with no errors? wouldnt stop on 1st error be simpler?

src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -58,22 +62,25 @@ public ConsumerRecordsProcessor(MessageAssembler messageAssembler,
...
+      long offset = record.offset();
+      try {
+        ConsumerRecord<K, V> handledRecord = handleConsumerRecord(record);
+        result.addRecord(tp, handledRecord);
+      } catch (RuntimeException e) {
+        if (!_skipRecordOnException) {
+          result.recordException(tp, offset, e);
         }
-        list.add(handledRecord);
+        LOG.warn("Exception thrown when processing message with offset {} from partition {}", offset, tp, e);
```
radai-rosenblatt: ,ove log.warn before recordException? (makes the log more coherent)

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
```diff
@@ -78,7 +78,7 @@ public MessageSplitterImpl(int maxSegmentSize,
     int messageSizeInBytes = serializedRecord.length;
     ByteBuffer bytebuffer = ByteBuffer.wrap(serializedRecord);
 
-    byte[] segmentKey = key == null ? LiKafkaClientsUtils.uuidToBytes(segmentMessageId) : key;
+    byte[] segmentKey = (key == null && numberOfSegments > 1) ? LiKafkaClientsUtils.uuidToBytes(segmentMessageId) : key;
```
radai-rosenblatt: not related to PR topic, right ?

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -221,6 +227,14 @@ public void unsubscribe() {
 
   @Override
   public ConsumerRecords<K, V> poll(long timeout) {
+    if (_lastProcessedResult != null && _lastProcessedResult.exception() != null) {
```
radai-rosenblatt: since we throw the exception from processing batch N-1 on the Nth call to poll(), wouldnt an "early" call to commit + crash before Nth poll() call make us skip an entire batch ?

smccauliff: Shouldn't offsets be modified after the call to the consumer records processor returns?  That would have similar semantics to the existing autocommit.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -221,6 +227,14 @@ public void unsubscribe() {
...
+      for (Map.Entry<TopicPartition, Long> entry : _lastProcessedResult.resumeOffsets().entrySet()) {
+        _kafkaConsumer.seek(entry.getKey(), entry.getValue());
+      }
+      RuntimeException e = _lastProcessedResult.exception();
+      _lastProcessedResult = null;
+      throw e;
```
smccauliff: I thought the point of patch was to not throw these exceptions?

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -261,7 +275,8 @@ public void unsubscribe() {
           }
         }
       }
-      processedRecords = _consumerRecordsProcessor.process(rawRecords);
+      _lastProcessedResult = _consumerRecordsProcessor.process(rawRecords);
```
smccauliff: _lastProcessedResult will still have all the processed records in it which will be retained in-memory until the next call to pool.  Since this code also deals with large messages this may be quite a bit of retained memory.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
becketqin: No, the point of the patch is double folded. If the `skip.record.on.exception` is set to true. We do not throw exception. Otherwise, we throw exception, but user can choose to catch it and continue to call poll() so the exception will be skipped.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
becketqin: Good point.

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
becketqin: Yes, that is probably better.

becketqin: Talked offline. This won't happen.

src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
becketqin: Right, this is to preserve the null key behavior so that it won't be accidentally set to a UUID.

src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
becketqin: We don't have to. This is to avoid refetching the messages. So we only stop if there is an exception in the partition.

PR #48, opened 2017-06-21
Add sanity check logic to the message eviction.

becketqin:
We saw a case that the used buffer recorded by LargeMessageBufferPool was none zero while there was no large message buffered. Adding the sanity check logic and hopefully we can catch the issue.


becketqin:
ping @radai-rosenblatt @smccauliff

PR #50, opened 2017-07-10
Enable producer and consumer to skip record on SkippableException

lindong28:


lindong28:
This patch made the following changes:

- Fixed bug in `LiKafkaProducerImpl` to avoid double counting the failure when there is produce failure.
- Added `SkippableException` which may be thrown from user-specified custom serializer/deserializer. For example, if serialize/de-serializer can not connect to schema registry due to IOException, serializer/deserializer can throw `SkippableException`.
- Added config SKIP_RECORD_ON_SKIPPABLE_EXCEPTION_CONFIG for both producer and consumer. `producer.send(..)` will return null if this config is true and there is a `SkippableException`. `consumer.poll()` will ignore record with `SkippableException`.

lindong28:
Ping @smccauliff @radai-rosenblatt @becketqin for review.

smccauliff:
Unit testing?

lindong28:
@smccauliff Ah.. my bad. The test is fixed now.

smccauliff:
I mean can you add a unit test that tests the enabled and disabled states?

lindong28:
@smccauliff Sure, I have added `LiKafkaProducerIntegrationTest.testSerializationException()` for producer. The existing `ConsumerRecordsProcessorTest.testDeserializationException` covers consumer. Thanks!

src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -80,12 +80,14 @@
       + "the offset to the latest offset</li><li>none: throw exception to the consumer if no previous offset is found "
       + "for the consumer's group</li><li>anything else: throw exception to the consumer.</li></ul>";
 
-  private static final String SKIP_RECORD_ON_EXCEPTION_DOC = "Skip the record with RecordProcessingException. "
-      + "This is to allow the users who do are willing to ignore the problematic record to continue consuming without "
+  private static final String SKIP_RECORD_ON_SKIPPABLE_EXCEPTION_DOC = "Skip the record with SkippableException. "
```
smccauliff: Might want to say something like:
"When processing a record and a SkippableException is thrown then a log message is generated and the exception is ignored.  The code will processed to process records as if the exception never happened.  The record which caused the SkippableException will be lost."

smccauliff: For some reason I thought this doc string was much shorter.  Never mind I like your original doc string.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -254,10 +256,18 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
...
       return future;
+    } catch (SkippableException e) {
+      _auditor.record(_auditor.auditToken(producerRecord.key(), producerRecord.value()), producerRecord.topic(),
+                      producerRecord.timestamp(), 1L, 0L, AuditType.FAILURE);
+      if (_skipRecordOnSkippableException) {
+        LOG.warn("Exception thrown when producing message to partition {}-{}", producerRecord.topic(), producerRecord.partition());
+        return null;
+      }
+      throw e;
     } catch (Throwable t) {
```
smccauliff: I thought we were going to top catching Throwable.

src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -254,10 +256,18 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
         future = _producer.send(wrappedRecord.get(0), errorLoggingCallback);
       }
       return future;
+    } catch (SkippableException e) {
+      _auditor.record(_auditor.auditToken(producerRecord.key(), producerRecord.value()), producerRecord.topic(),
```
smccauliff: The failure recording code can happen just once in a finally block.  No?

smccauliff:
I don't have anymore comments.

PR #52, opened 2017-07-14
Auditor should only call join() if timeout > 0

lindong28:


lindong28:
Thread.join(timeout) will wait forever if timeout is 0. This patch fixes the problem.

PR #58, opened 2017-07-31
Add constructor that allows injecting a Consumer

ambroff:
I'm working on a change to a library that uses LiKafkaConsumerImpl and I
found myself wanting to add an integration test that tested that
integration. I wanted to do that by providing a mock Consumer<byte[],
byte[]> so that my integration test didn't actually need Kafka, but
unfortunately there wasn't any way to inject a consumer object.

This change just adds a couple of new constructors so that clients can
provide their own Consumer<byte[], byte[]> if necessary.


becketqin:
Ideally we don't want to expose such injection as a public interface for testing purpose. I am wondering what was the test trying to do?

ambroff:
It's not just for testing purposes. Someone might want to provide their own custom implementation for some other reason. I see no downside to having these additional constructors in the interface.

PR #60, opened 2017-09-07
Add release on tag for circleci

radai-rosenblatt:
docs - https://circleci.com/docs/2.0/workflows/#git-tag-job-execution


smccauliff:
LGTM.  The only thing that you might want to add is some check that the version gradle knows about is also the version used in the tag.  Since this in the gradle.properties file it should be straight forward to extract this from gradle.properties.

radai-rosenblatt:
now with 60% more regexp

PR #65, opened 2017-09-20
abolish scala. make test harnesses java, avoid any link-time dependency on scala

radai-rosenblatt:


src/main/java/com/linkedin/kafka/clients/producer/UUIDFactory.java
```diff
@@ -17,12 +17,13 @@
 public interface UUIDFactory<K, V> extends Configurable {
 
   /**
-   * Get a random UUID.
+   * @return a random UUID
```
smccauliff: If you are cleaning up documentation then you might as well change this to something like "a non-null UUID".  The UUID does not need to be random.

src/test/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerIntegrationTest.java
```diff
@@ -1052,6 +1055,14 @@ private void produceSyntheticMessages(String topic) {
     producer.close();
   }
 
+  //TODO - remove / refactor
+  private Producer<byte[], byte[]> createKafkaProducer() {
```
smccauliff: Should we default all tests to SSL?

src/test/java/com/linkedin/kafka/clients/utils/tests/AbstractKafkaClientsIntegrationTestHarness.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public void setUp() {
+    super.setUp();
+  }
+
+  @Override
+  public void tearDown() {
+    try {
+      //TODO - something here ?
```
smccauliff: Eh?

src/test/java/com/linkedin/kafka/clients/utils/tests/AbstractKafkaClientsIntegrationTestHarness.java
```diff
@@ -0,0 +1,113 @@
...
+
+    result.setProperty(LiKafkaConsumerConfig.MESSAGE_ASSEMBLER_BUFFER_CAPACITY_CONFIG, "300000");
+    result.setProperty(LiKafkaConsumerConfig.MESSAGE_ASSEMBLER_EXPIRATION_OFFSET_GAP_CONFIG, "10000");
+    result.setProperty(LiKafkaConsumerConfig.EXCEPTION_ON_MESSAGE_DROPPED_CONFIG, "true");
+    result.setProperty(LiKafkaConsumerConfig.MAX_TRACKED_MESSAGES_PER_PARTITION_CONFIG, "10000");
+    result.setProperty(LiKafkaConsumerConfig.SEGMENT_DESERIALIZER_CLASS_CONFIG, DefaultSegmentDeserializer.class.getCanonicalName());
+
+    //apply overrides
+    if (overrides != null) {
+      overrides.forEach(result::put);
```
smccauliff: result.putAll(overrides)?

src/test/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerIntegrationTest.java
radai-rosenblatt: no. this method is rather hacky and creates an ad-hoc producer<byte[], byte[]> where the rest of the harness code creates <string, string>, and so should be cleaned up somehow.
however, our tests currently rely on this heavily nd i didnt want ot make this PR any bogger.

src/test/java/com/linkedin/kafka/clients/utils/tests/AbstractKafkaIntegrationTestHarness.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return broker;
+  }
+
+  public String bootstrapServers() {
+    return bootstrapUrl;
+  }
+
+
+  protected List<Map<Object, Object>> buildBrokerConfigs() {
```
smccauliff: It might be useful to add documentation about this return value so the user knows that returnValue.get(i) is a config for the ith broker.

src/test/java/com/linkedin/kafka/clients/utils/tests/AbstractZookeeperTestHarness.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").  See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.utils.tests;
+
+public abstract class AbstractZookeeperTestHarness {
+  protected EmbeddedZookeeper zookeeper = null;
```
smccauliff: zookeeper -> _zookeeper

src/test/java/com/linkedin/kafka/clients/utils/tests/KafkaTestUtils.java
```diff
@@ -0,0 +1,124 @@
...
+  static {
+    SHUTDOWN_HOOK = new Thread(() -> {
+      for (File toCleanUp : FILES_TO_CLEAN_UP) {
+        if (!toCleanUp.exists()) {
+          continue;
+        }
+        try {
+          FileUtils.forceDelete(toCleanUp);
+        } catch (IOException issue) {
+          issue.printStackTrace(System.err);
```
becketqin: Maybe rethrow the exception to let the uncaught exception handler handle it.

src/test/java/com/linkedin/kafka/clients/utils/tests/KafkaTestUtils.java
```diff
@@ -0,0 +1,124 @@
...
+    props.put("auto.offset.reset", "earliest");
+    props.put("enable.auto.commit", "false");
+    props.put("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+    props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+    KafkaConsumer<String, String> consumer = new KafkaConsumer<>(props);
+
+    return consumer;
+  }
+
+  public static int getAvailableTcpPort() {
```
becketqin: Can we reuse the `KafkaServer.boundPort()` method and `ZookeeperServer.getClientPort()` instead of writing our own method?

src/test/java/com/linkedin/kafka/clients/utils/tests/AbstractKafkaIntegrationTestHarness.java
```diff
@@ -0,0 +1,121 @@
...
+      builder.zkConnect(zookeeper());
+      builder.nodeId(i);
+      builder.enable(securityProtocol());
+      if (securityProtocol() == SecurityProtocol.SSL) {
+        if (trustStoreFile() != null) {
+          builder.trustStore(trustStoreFile());
+        }
+      } else {
+        if (trustStoreFile() != null) {
+          throw new AssertionError("security protocol not yet yet trust store file provided");
```
smccauliff: I don't understand this error message.

src/test/java/com/linkedin/kafka/clients/utils/tests/EmbeddedBroker.java
```diff
@@ -0,0 +1,145 @@
...
+  private final static Class<?> SCALA_SEQ_CLASS;
+  private final static Class<?> SCALA_ARRAY_BUFFER_CLASS;
+  private final static Constructor<?> CONFIG_CTR;
+  private final static Constructor<?> SERVER_CTR;
+  private final static Constructor<?> ARR_BUF_CTR;
+  private final static Method STARTUP_METHOD;
+  private final static Method SHUTDOWN_METHOD;
+  private final static Method AWAIT_SHUTDOWN_METHOD;
+  private final static Object EMPTY_OPTION;
+  private final static Pattern LISTENER_PATTERN = Pattern.compile("\\s*(\\w+)://([\\w\\d]+):(\\d+)\\s*");
```
smccauliff: This won't match IPv6 addresses which someone might want to use if they were testing that IPv6 worked.

smccauliff: See https://www.ietf.org/rfc/rfc2732.txt

smccauliff: Probably the \\s* at the beginning and end is not needed.

src/test/java/com/linkedin/kafka/clients/utils/tests/EmbeddedBrokerBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+          }
+          break;
+        default:
+          throw new UnsupportedOperationException("unhandled: " + interBrokerSecurity);
+      }
+    }
+    return this;
+  }
+
+  public EmbeddedBrokerBuilder trustStore(File trustStore) {
```
smccauliff: Do the trust stores need to be files or can they be resources so that you can package truststores in the source code.

src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -50,7 +47,7 @@ public int clusterSize() {
   @Override
   public Properties overridingProps() {
     Properties props = new Properties();
-    props.setProperty(KafkaConfig.NumPartitionsProp(), Integer.toString(NUM_PARTITIONS));
+    props.setProperty("num.partitions", Integer.toString(NUM_PARTITIONS));
```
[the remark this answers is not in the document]
radai-rosenblatt: that prop is in scala code. it is unclean ...

src/test/java/com/linkedin/kafka/clients/utils/tests/KafkaTestUtils.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.kafka.clients.producer.KafkaProducer;
+
+
+public class KafkaTestUtils {
+  private final static AtomicBoolean SHUTDOWN_HOOK_INSTALLED = new AtomicBoolean(false);
+  private final static Thread SHUTDOWN_HOOK;
+  private final static List<File> FILES_TO_CLEAN_UP = Collections.synchronizedList(new ArrayList<>());
+
+  static {
+    SHUTDOWN_HOOK = new Thread(() -> {
```
smccauliff: Is this just here just in case tear down does not work?

src/test/java/com/linkedin/kafka/clients/utils/tests/EmbeddedBrokerBuilder.java
radai-rosenblatt: the scala harness uses a File as a trust store, and i matched it.
also, this is test code, i see no use case for packaging trust stores into the artifacts ?

src/test/java/com/linkedin/kafka/clients/utils/tests/KafkaTestUtils.java
radai-rosenblatt: no. closable test classes (like embeddable brokers and zk) should delete their files on close, but other tests may need files/folders

PR #68, opened 2017-09-22
Fix the issue that NPE is thrown when user sends null value.

becketqin:


src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
```diff
@@ -65,10 +66,12 @@ public MessageSplitterImpl(int maxSegmentSize,
                                                     byte[] key,
                                                     byte[] serializedRecord,
                                                     int maxSegmentSize) {
-
     if (topic == null) {
       throw new IllegalArgumentException("Topic cannot be empty for LiKafkaGenericMessageSplitter.");
     }
+    if (serializedRecord == null) {
+      return Collections.singletonList(new ProducerRecord<>(topic, partition, timestamp, key, null));
```
abhishekmendhekar: This line returns an immutable list where are the `segments` returns a mutable list [here](https://github.com/linkedin/li-apache-kafka-clients/pull/68/files#diff-fea16f184ae9e84090ff9fdf721e695eR77). Hopefully it shouldn't break any thing for callers of this method.

becketqin: Thanks @abhishekmendhekar. The List is not supposed to be changed. So we should be fine here.

PR #79, opened 2017-11-14
split into multiple modules

radai-rosenblatt:
this splits the code into 3 modules:

1. li-apache-kafka-clients - main codebase
2. kafka-test-harness - contains the test harness code. depends on 1
3. integration-tests - uses the test harness code (2) to test 1

we expect to publish 1 and 2 as artifacts. 3 is our own integration tests and had to be factored out to avoid a circular dependency, but shouldnt be used by anyone else.


mockitoguy:
Can you write a short description what is driving the changes?

This will help me review the changes! Thank you!

becketqin:
@radai-rosenblatt Thanks for the patch. LGTM. The only comment I have is that it might be better to keep the idea plugin so people who are familiar with that can still use it.

@mockitoguy It's great to have some expert to look at this :)

radai-rosenblatt:
@mockitoguy - i've updated the commit msg. here is the new one:

split to multiple modules
    
    this codebase contains 2 usable artifats already - the clients, and the test harness base classes, which before this patch were included in a test artifact.
    a test artifact is awkward to produce, publish and depend upon so this splits the project up into li-apache-kafka-clients (which contains the clients, as t always has)
    and kafka-test-harness, which holds the base classes for the test harness. these classes are useful in starting/stopping embedded ZK and kafka for integration tests.
    there's also a 3rd module which uses the test harness to run integration tests for the clients. this 3rd module does not produce any artifact for downstream users.

hopefully this explains the changes and the reasoning behind them

buildSrc/src/main/groovy/com/linkedin/gradle/build/DistributeTask.groovy
```diff
@@ -9,46 +9,46 @@ import org.jfrog.gradle.plugin.artifactory.dsl.ArtifactoryPluginConvention
 
 class DistributeTask extends DefaultTask {
```
mockitoguy: I would not roll out buildSrc for just one class. I would drop it in a separate gradle file and import the file using "apply from:". However, it's a nitpick your code works ;)

DistributeTask.groovy is Groovy code but has java artifacts like semicolons

li-apache-kafka-clients/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+dependencies {
+    compile "org.apache.kafka:kafka-clients:${rootProject.ext.kafkaVersion}"
+}
+
+publishing {
+    publications {
+        java(MavenPublication) {
+            from components.java
```
mockitoguy: Minor duplication that you can wrap if you push out the common logic to script plugin applied with "apply from:"

settings.gradle
```diff
@@ -1,6 +1,24 @@
...
+    classpath 'gradle.plugin.net.vivin:gradle-semantic-build-versioning:4.0.0'
+  }
+}
+
+apply plugin: 'net.vivin.gradle-semantic-build-versioning'
+
+//otherwise it defaults to the folder name
 rootProject.name = 'li-apache-kafka-clients'
 
+include 'li-apache-kafka-clients'
```
mockitoguy: This is great, nicely modular now!

FYI, you can include in one line: include 'x', 'y', etc.

.circleci/config.yml
```diff
@@ -36,9 +36,7 @@ jobs:
        - restore_cache:
            key: dependency-cache-{{ checksum "build.gradle" }}
        - run:
-           command: ./gradlew assemble testJar
-       - run:
-           command: ./.circleci/publish.sh
+           command: ./gradlew distributeBuild
```
mockitoguy: Not running tests before the release? distributeBuild does not trigger running the checks.

build.gradle
```diff
@@ -3,178 +3,130 @@
...
+    artifacts {
+      archives sourcesJar
+      archives javadocJar
     }
+  }
 }
 
-//publishing (artifactory)
-
+artifactoryPublish.skip = true
```
mockitoguy: Does it mean that the automated publications are disabled. Is it intentional?

.circleci/config.yml
radai-rosenblatt: in config.yml the publish job requires (a passing) build, so tests would get run before publication

build.gradle
radai-rosenblatt: i think this just disables publishing the parent project. im not sure as ethan gave me this part of the conf

li-apache-kafka-clients/build.gradle
radai-rosenblatt: beyond my gradle knowledge. would you be ok with submitting a follow-up PR ?

buildSrc/src/main/groovy/com/linkedin/gradle/build/DistributeTask.groovy
radai-rosenblatt: removed the semi colons. i'd be happiest to not have this at all, but that depends on https://github.com/JFrogDev/build-info/pull/128 happening.

lindong28:
@radai-rosenblatt @mockitoguy It seems that this commit breaks `gradle idea` with the following error:

```
* What went wrong:
A problem occurred evaluating root project 'buildSrc'.
> No signature of method: build_4kafg6gltq7e3kb5renv2ehpm8$_run_closure1.id() is applicable for argument types: (java.lang.String) values: [java-gradle-plugin]
  Possible solutions: is(java.lang.Object), is(java.lang.Object), find(), find(), find(groovy.lang.Closure), find(groovy.lang.Closure)
```

I am not able to fix it myself. Do you know how to fix it?

PR #83, opened 2017-12-20
Fix the case that position throws exception on a good partition due t…

becketqin:
…o batched query of positions.


lindong28:
LGTM

PR #91, opened 2018-03-05
fix bug where committed offsets were not making progress when all msgs are too large

radai-rosenblatt:
code was check-pointing the offset(s) of the last msg that was ever delivered. there may never have been such a msg...

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
@@ -8,6 +8,7 @@ import com.linkedin.kafka.clients.auditing.AuditType;
 import com.linkedin.kafka.clients.auditing.Auditor;
 import com.linkedin.kafka.clients.largemessage.errors.SkippableException;
 import com.linkedin.kafka.clients.utils.LiKafkaClientsUtils;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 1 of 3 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -172,12 +173,19 @@ public Long safeOffset(TopicPartition tp, long messageOffset) {
...
+          //any hope of delivering a large message (maybe we dropped a lot of them due to low memory)
+          safeOffsetToCommit = _messageAssembler.safeOffset(tp);
         } else {
-          //In this case nothing has been delivered to the consumer (this this partition) either because we never
-          // reached the highwatermark or because we were waiting to assemble some message
-          safeOffsetToCommit = _deliveredMessageOffsetTracker.safeOffset(tp);
+          if (previousDeliveredOffset != null) {
+            safeOffsetToCommit = _deliveredMessageOffsetTracker.safeOffset(tp, previousDeliveredOffset);
+          } else {
+            safeOffsetToCommit = _messageAssembler.safeOffset(tp);
```
smccauliff: What does this do?

radai-rosenblatt: deliveredMessageOffsetTracker has history, so can map "historical" offsets delivered to users to their "real" underlying offsets.

messageAssembler only deals with the current stats.

this code means that if we have history (because we've delivered some != 0 msgs before, we use deliveredMessageOffsetTracker to get the real offset, otherwise use messageAssembler

becketqin:
LGTM.

PR #99, opened 2018-10-29
Porting all fixes from linkinedin. 

navina:
Major ones are:
-  Add bounded flush to LiKafkaProducer
- Expire large message segments when normal messages arrive.
-  LICLOSEST implementation & integration tests (introduced `LiOffsetResetStrategy` to avoid using `String`) 
- Don't retrieve the consumer high watermark for every partition on every offset commit.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
@@ -113,6 +113,7 @@ import java.util.concurrent.atomic.AtomicInteger;
  */
 public class LiKafkaProducerImpl<K, V> implements LiKafkaProducer<K, V> {
   private static final Logger LOG = LoggerFactory.getLogger(LiKafkaProducerImpl.class);
+  private static final String BOUNDED_FLUSH_THREAD_PREFIX = "Bounded-Flush-Thread-";
 
   // Large message settings
   private final boolean _largeMessageEnabled;
@@ -133,7 +134,9 @@ public class LiKafkaProducerImpl<K, V> implements LiKafkaProducer<K, V> {
   private final AtomicInteger _numThreadsInSend;
   private volatile boolean _closed;
 
-  private final Method _producerSupportsBoundedFlush;
+  // This is null if the underlying producer does not have an implementation for time-bounded flush
+  private final Method _boundedFlushMethod;
+  private final AtomicInteger _boundFlushThreadCount = new AtomicInteger();
 
   public LiKafkaProducerImpl(Properties props) {
     this(new LiKafkaProducerConfig(props), null, null, null, null);
@@ -172,10 +175,10 @@ public class LiKafkaProducerImpl<K, V> implements LiKafkaProducer<K, V> {
     try {
       producerSupportsBoundedFlush = _producer.getClass().getMethod("flush", long.class, TimeUnit.class);
     } catch (NoSuchMethodException e) {
-      LOG.warn("LiKafkaProducerImpl does not support bounded flush.");
+      LOG.warn("Wrapped KafkaProducer does not support time-bounded flush.", e);
       producerSupportsBoundedFlush = null;
     }
-    _producerSupportsBoundedFlush = producerSupportsBoundedFlush;
+    _boundedFlushMethod = producerSupportsBoundedFlush;
 
     try {
       // Instantiate the key serializer if necessary.
@@ -306,25 +309,29 @@ public class LiKafkaProducerImpl<K, V> implements LiKafkaProducer<K, V> {
    */
   @Override
   public void flush(long timeout, TimeUnit timeUnit) {
-    boolean invokeFlush = false;
-    if (_producerSupportsBoundedFlush != null) {
+    boolean useSeparateThreadForFlush = false;
+    if (_boundedFlushMethod != null) {
       try {
-        _producerSupportsBoundedFlush.invoke(_producer, timeout, timeUnit);
+        _boundedFlushMethod.invoke(_producer, timeout, timeUnit);
       } catch (IllegalAccessException | InvocationTargetException e) {
         LOG.trace("Underlying producer does not support bounded flush!", e);
-        invokeFlush = true;
+        useSeparateThreadForFlush = true;
       }
     } else {
-      invokeFlush = true;
+      useSeparateThreadForFlush = true;
     }
 
-    if (invokeFlush) {
+    if (useSeparateThreadForFlush) {
       final CountDownLatch latch = new CountDownLatch(1);
       Thread t = new Thread(() -> {
         _producer.flush();
         latch.countDown();
       });
       t.setDaemon(true);
+      t.setName(BOUNDED_FLUSH_THREAD_PREFIX + _boundFlushThreadCount.getAndIncrement());
+      t.setUncaughtExceptionHandler((t1, e) -> {
+        LOG.warn("Thread " + t1.getName() + " terminated unexpectedly.", e);
+      });
       t.start();
 
       boolean latchResult = false;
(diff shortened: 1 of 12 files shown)


smccauliff:
I spot checked a couple of things to make sure the code that I think should be there is still there. I edited the PR description.  Is there something I should look at specifically?  Are there any changes from our internal repo other than the new enum?

navina:
> I spot checked a couple of things to make sure the code that I think should be there is still there. I edited the PR description. Is there something I should look at specifically? Are there any changes from our internal repo other than the new enum?

Yes. The new enum is different from internal repo. The other thing that is different is how we are addressing bounded flush in `LiKafkaProducerImpl`. The new flush API is pending commit in apache kafka. Hence, as a workaround to guarantee a "bounded" flush, we use reflection and if needed, invoke unbounded flush in a separate thread. It is nasty hack. But it maintains the user API guarantees. So,  implementation of`LiKafkaProducerImpl#flush(long,TimeUnit)` is the other major change.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -128,6 +133,8 @@
   private final AtomicInteger _numThreadsInSend;
   private volatile boolean _closed;
 
+  private final Method _producerSupportsBoundedFlush;
```
smccauliff: Can this be _boundedFlushMethod with a comment /* this is null if the producer does not have a method for time bounded flush. */

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -160,6 +167,15 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
                              Auditor<K, V> auditor) {
     // Instantiate the open source producer, which always sents raw bytes.
     _producer = new KafkaProducer<>(configs.originals(), new ByteArraySerializer(), new ByteArraySerializer());
+    // TODO: This hack remains until bounded flush is added to apache/kafka
+    Method producerSupportsBoundedFlush;
+    try {
+      producerSupportsBoundedFlush = _producer.getClass().getMethod("flush", long.class, TimeUnit.class);
+    } catch (NoSuchMethodException e) {
+      LOG.warn("LiKafkaProducerImpl does not support bounded flush.");
```
smccauliff: Maybe the message should say "Wrapped Kafka producer does not have a time bounded flush method."

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -284,6 +300,45 @@ public void flush() {
     _producer.flush();
   }
 
+  /**
+   * This method will flush all the message buffered in producer. The call blocks until timeout.
+   * If the underlying producer doesn't support a bounded flush, it will invoke the {@link #flush()}.
+   */
+  @Override
+  public void flush(long timeout, TimeUnit timeUnit) {
+    boolean invokeFlush = false;
```
smccauliff: invokeFlush -> useSeparateThreadForFlush ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -284,6 +300,45 @@ public void flush() {
     _producer.flush();
   }
 
+  /**
+   * This method will flush all the message buffered in producer. The call blocks until timeout.
+   * If the underlying producer doesn't support a bounded flush, it will invoke the {@link #flush()}.
+   */
+  @Override
+  public void flush(long timeout, TimeUnit timeUnit) {
```
smccauliff: Should this be synchronized?  If more than one thread calls flush then all of them may need to spin up an extra thread.

smccauliff: Probably this needs to see if the producer is already closed.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
radai-rosenblatt: i'd actually print the expected method signature not found

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -284,6 +300,45 @@ public void flush() {
...
+        invokeFlush = true;
+      }
+    } else {
+      invokeFlush = true;
+    }
+
+    if (invokeFlush) {
+      final CountDownLatch latch = new CountDownLatch(1);
+      Thread t = new Thread(() -> {
+        _producer.flush();
```
radai-rosenblatt: i'd name the thread, just in case

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
radai-rosenblatt: i dont think we should be trying to provide any more concurrency control than the underlying client

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
radai-rosenblatt: and add an uncaught exception handler that prints them to log

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
navina: Sure. Makes sense.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
navina: Ok I will add the method signature.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
navina: sure

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
navina: @smccauliff  If multiple thread call `flush` today, we don't handle it with any additional synchronization than what the underlying producer provides. So, why should that be a concern here? Is it because of the number of threads we end up creating?

navina: @smccauliff is the unbounded number of thread creation your only concern here? We can overcome that by limiting the number of threads we spin up and block when the ceiling is reached. This changes the behavior of the flush method as it exists today. Let me know what you think? 
As far as this version is concerned, we don't want to do any incompatible changes.

smccauliff: I remember discussing this with you in person several weeks ago and I think we concluded that your implementation was the correct path.

smccauliff: Let me take another look tomorrow when I'm more awake.

navina: We discussed? I don't remember :( Take a look when you can.

PR #102, opened 2019-01-02
Make topic name accessible in the MessageSplitterImpl

navina:
Even though we are passing topic here to serialize method, the segment itself should be topic independent. This fix was missed when porting all changes from Linkedin.


jonlee2:
LGTM. Thanks for the fix.

PR #106, opened 2019-02-15
Federated Clients: implement LiKafkaFederatedProducerImpl#close() and flush()

jonlee2:


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -105,6 +107,10 @@ public LongAdder numThreadsInSend() {
     public boolean closed() {
```
navina: Why do we need a per producer context `closed()` public API ? It looks like you are currently using it only for tests.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -355,7 +361,68 @@ public void close() {
...
+    List<Thread> taskThreads = new ArrayList<>();
+    for (Map.Entry<ClusterDescriptor, PerClusterProducerContext<K, V>> entry : _producers.entrySet()) {
+      Thread taskThread = new Thread(new CloseTask(entry.getValue(), deadlineTimeMs));
+      taskThreads.add(taskThread);
+      taskThread.start();
+    }
+
+    for (Thread taskThread : taskThreads) {
+      try {
+        taskThread.join();
```
navina: Would a countdown latch or cyclic barrier work better here? the for-loop imposes some kind of implicit order in which the threads terminate

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -355,7 +361,68 @@ public void close() {
...
+    private long _deadlineTimeMs;
+
+    CloseTask(PerClusterProducerContext<K, V> context, long deadlineTimeMs) {
+      _context = context;
+      _deadlineTimeMs = _deadlineTimeMs;
+    }
+
+    public void run() {
+      LOG.info("Closing LiKafkaProducer for cluster {}", _context.cluster());
+      LongAdder numThreadsInSend = _context.numThreadsInSend();
```
navina: I want to understand the association between the number of producers per cluster (I am assuming cluster here is actually a cluster group) and the number of threads in Sends. I am not clear on why we need a `numThreadsInSend` when there is a similar logic in `LiKafkaProducerImpl` as well.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -355,7 +361,68 @@ public void close() {
...
+
+      _context.setClosed(true);
+      synchronized (numThreadsInSend) {
```
navina: I am not sure why we need synchronize here if `numThreadsInSend` is per-producer context?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: It is used for LiKafkaFederatedProducerImpl as well.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: I think functionality-wise, using join should be fine (each thread is supposed to be independent and I don't think implicit ordering doesn't really matter). But I switched to countdown latch for better readability/simplicity.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: The "cluster" here is actually a physical cluster, not a cluster group, and we maintain a separate numThreadsInSend per physical cluster via context.

I copied the send() implementation from LiKafkaProducerImpl here. I am not calling LiKafkaProducerImpl#send().

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: I think we should put notifyAll() in a synchronized block, otherwise a notification may be lost due to a race condition.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: Per our discussion, I restructured the code to directly create and use per-cluster LiKafkaProducer.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -385,29 +243,27 @@ public void abortTransaction() throws ProducerFencedException {
...
     // Create per-cluster producer config with the actual bootstrap URL of the physical cluster to connect to.
     Map<String, Object> configMap = _commonProducerConfigs.originals();
     configMap.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, cluster.bootstrapURL());
-    LiKafkaProducerConfig perClusterProducerConfig = new LiKafkaProducerConfig(configMap);
-    _producerBuilder.setProducerConfig(perClusterProducerConfig);
-    PerClusterProducerContext newContext = new PerClusterProducerContext(cluster, perClusterProducerConfig,
-        _producerBuilder.build());
-    _producers.put(cluster, newContext);
-    return newContext;
+    LOG.error("##### producer Config: " + configMap);
```
navina: Why error?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -355,7 +185,35 @@ public void close() {
...
+      return;
+    }
+
+    LOG.info("Closing LiKafkaProducer for cluster group {} in {} {}...", _clusterGroup, timeout, timeUnit);
+
+    long startTimeMs = System.currentTimeMillis();
+    long deadlineTimeMs = startTimeMs + timeUnit.toMillis(timeout);
+    CountDownLatch countDownLatch = new CountDownLatch(_producers.entrySet().size());
+    for (Map.Entry<ClusterDescriptor, LiKafkaProducer<K, V>> entry : _producers.entrySet()) {
+      Thread taskThread = new Thread(() -> {
```
navina: nit: you could get away without a local variable

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -355,7 +185,35 @@ public void close() {
...
-    throw new UnsupportedOperationException("Not implemented yet");
+    if (_producers.isEmpty()) {
+      LOG.warn("No producers to close for cluster group {}", _clusterGroup);
+      return;
+    }
+
+    LOG.info("Closing LiKafkaProducer for cluster group {} in {} {}...", _clusterGroup, timeout, timeUnit);
+
+    long startTimeMs = System.currentTimeMillis();
+    long deadlineTimeMs = startTimeMs + timeUnit.toMillis(timeout);
```
navina: Possible overflow: when `close()` is invoked, the timeout value is `Integer.MAX_VALUE`. Is it safe to compute deadline as `startTimeMs + timeUnit.toMillis(timeout)` ? Fwiw, I understand this is similar in the existing `LiKafkaProducerImpl`.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
```diff
@@ -328,13 +131,40 @@ public void flush() {
...
-    throw new UnsupportedOperationException("Not implemented yet");
+    if (_producers.isEmpty()) {
+      LOG.warn("No producers to flush for cluster group {}", _clusterGroup);
+      return;
+    }
+
+    LOG.info("Flushing LiKafkaProducer for cluster group {} in {} {}...", _clusterGroup, timeout, timeUnit);
+
+    long startTimeMs = System.currentTimeMillis();
+    long deadlineTimeMs = startTimeMs + timeUnit.toMillis(timeout);
```
navina: Ah.. flush is also doing the same thing with MAX_VALUE ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").
 See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.producer;
+
+import java.util.Map;
+
+
+public class LiKafkaProducerBuilder<K, V> {
```
navina: Being a public class, it is not clear who should be using this - is it only meant for users of LiFederatedKafkaClients or is it also meant for users of LiKafkaProducer ? Add some documentations to clarify. if we don't want a general usage of this class, we should change it to a non-public scope.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: oops.. forgot to remove... removed.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: Done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: Given that the timeout is Integer.MAX_VALUE and we are using long for the deadline, I think it should be safe (at least as long as we live.. :P). I admit that this is not the most elegant way, though. I also noticed the open source Kafka code also does this.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaFederatedProducerImpl.java
jonlee2: See the above comment.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerBuilder.java
jonlee2: This is meant for users of federate clients only. I don't really see users (other than tests) to provide this when constructing federated clients, so I think it makes sense to make it package private.

PR #107, opened 2019-02-20
Various improvements to LiKafkaProducerImpl.

jonlee2:
1. Use LongAdder instead of AtomicInteger, which is more performant.
2. Use the large.message.segment.wrapping.required property to control whether or not wrap a normal-sized message in a large message segment
  (The default value is set to false for now, to match the previous behavior.)
3. Use the min of max.message.segment.bytes and max.request.size for the max message segment size for large message.
4. Use an exception instead of an assert

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
@@ -90,7 +90,7 @@ public class LiKafkaProducerConfig extends AbstractConfig {
         .define(CLUSTER_GROUP_CONFIG, Type.STRING, "", Importance.MEDIUM, CLUSTER_GROUP_DOC)
         .define(CLUSTER_ENVIRONMENT_CONFIG, Type.STRING, "", Importance.MEDIUM, CLUSTER_ENVIRONMENT_DOC)
         .define(MAX_REQUEST_SIZE_CONFIG, Type.INT, 1 * 1024 * 1024, atLeast(0), Importance.MEDIUM, MAX_REQUEST_SIZE_DOC)
-        .define(LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_CONFIG, Type.BOOLEAN, "false", Importance.MEDIUM,
+        .define(LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_CONFIG, Type.BOOLEAN, "true", Importance.MEDIUM,
             LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_DOC);
   }
 
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
@@ -40,7 +40,7 @@ import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.LongAdder;
 
 /**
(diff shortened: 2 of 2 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
```diff
@@ -90,7 +90,7 @@
         .define(CLUSTER_GROUP_CONFIG, Type.STRING, "", Importance.MEDIUM, CLUSTER_GROUP_DOC)
         .define(CLUSTER_ENVIRONMENT_CONFIG, Type.STRING, "", Importance.MEDIUM, CLUSTER_ENVIRONMENT_DOC)
         .define(MAX_REQUEST_SIZE_CONFIG, Type.INT, 1 * 1024 * 1024, atLeast(0), Importance.MEDIUM, MAX_REQUEST_SIZE_DOC)
-        .define(LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_CONFIG, Type.BOOLEAN, "false", Importance.MEDIUM,
+        .define(LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_CONFIG, Type.BOOLEAN, "true", Importance.MEDIUM,
```
navina: Changing the default value is dangerous. Do we have a new release after adding this config value?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -131,12 +132,12 @@
...
   // A counter of the threads in the middle of sending messages. This is needed to ensure when we close the producer
   // everything is audited.
-  private final AtomicInteger _numThreadsInSend;
+  private final LongAdder _numThreadsInSend = new LongAdder();
   private volatile boolean _closed;
 
   // This is null if the underlying producer does not have an implementation for time-bounded flush
   private final Method _boundedFlushMethod;
-  private final AtomicInteger _boundFlushThreadCount = new AtomicInteger();
+  private final LongAdder _boundFlushThreadCount = new LongAdder();
```
navina: This patch replaces `AtomicInteger` with `LongAdder` for `_numThreadsInSend` and `_boundFlushThreadCount`. I can see how there might be some higher contention for `_numThreadsInSend` and that it might benefit from switching to a `LongAdder` type. I don't think we can say the same for `_boundFlushThreadCount` because:
- It is unlikely that there are many user threads concurrently trying to flush the producer. 
- It seems like we are reading the `intValue` of the long adder `_boundFlushThreadCount` to create a unique identifier for the flush thread. I think that is pretty expensive and not guaranteed to return accurate result under concurrent updates https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/LongAdder.html#sum--

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
jonlee2: This is a new property that was practically never used (accessed by federated producer only), so I think it should be okay.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
jonlee2: Good point on LongAdder not providing an atomic snapshot. I agree that it should not be used for unique number generation. Javadoc on LongAdder does say that it is preferred for updating a common sum, but not for fine-grained synchronization control. I will revert both to AtomicInteger for now.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
navina: Ok

PR #110, opened 2019-03-07
fix message loss between poll() and close()

abhishekmendhekar:
poll() keeps a cache of exceptions per partition and throws exception in subsequent poll() call.
poll() also seek offset past exception assuming next poll() will throw exception.
This can lead to message loss if the user does not call poll() and commits offsets.

The fix stores the current offset and resume offset along with the exception. On exception,
it seeks to current offset (before exception) along with resume offset (current + 1).
On next poll(), poll() will seek to resume offset and throw exception and the user can decide
on either ignoring it which will skip the offset or exit.
On next seek*(), will throw an IllegalStateException and seeking to resume offsets, leaving the
decision to user to continue poll() or crash.
On exit, the user can safely commit offsets it consumed (excluding the offset that hit an exception)
without losing messages.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
@@ -241,17 +241,23 @@ public class LiKafkaConsumerImpl<K, V> implements LiKafkaConsumer<K, V> {
     long now = System.currentTimeMillis();
     long deadline = now + timeout;
     do {
-      ConsumerRecordsProcessingException e = handleRecordProcessingException(null);
-      if (e != null) {
-        throw e;
+      ConsumerRecordsProcessingException crpe;
+
+      // throw exception to user if the current active (un-paused) topic-partitions has exceptions
+      Set<TopicPartition> unPausedTopicPartitions = new HashSet<>(_kafkaConsumer.assignment());
+      unPausedTopicPartitions.removeAll(_kafkaConsumer.paused());
+      crpe = handleRecordProcessingException(unPausedTopicPartitions);
+      if (crpe != null) {
+        throw crpe;
       }
+
       if (_autoCommitEnabled && now > _lastAutoCommitMs + _autoCommitInterval) {
         commitAsync();
         _lastAutoCommitMs = now;
       }
       ConsumerRecords<byte[], byte[]> rawRecords = ConsumerRecords.empty();
       try {
-         rawRecords = _kafkaConsumer.poll(deadline - now);
+         rawRecords = _kafkaConsumer.poll(Duration.ofMillis(deadline - now));
       } catch (OffsetOutOfRangeException | NoOffsetForPartitionException oe) {
         handleInvalidOffsetException(oe);
       }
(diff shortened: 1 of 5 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessResult.java
```diff
@@ -46,14 +56,14 @@ void recordException(TopicPartition tp, long offset, RuntimeException e) {
...
+    _offsetPair.putIfAbsent(tp, new OffsetPair(offset, offset + 1));
   }
 
   public void clearRecords() {
     _processedRecords = null;
   }
 
-  boolean hasError(TopicPartition tp) {
+  public boolean hasError(TopicPartition tp) {
     return resumeOffsets().containsKey(tp);
```
smccauliff: This is kind of expensive as resumeOffsets() constructs a new map each time this check is performed.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -461,6 +484,42 @@ private long positionMain(TopicPartition partition, Duration timeout) {
...
+        ConsumerRecordsProcessingException e = _lastProcessedResult.exception();
+        _lastProcessedResult = null;
+        return e;
+      }
+
+    }
+    return null;
+  }
+
+  private void seekToResumeOffsetsOnRecordProcessingExceptions() {
```
radai-rosenblatt: any other invocations of this method except the above one? if not the condition check could be avoided

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessResult.java
abhishekmendhekar:
```
public Map<TopicPartition, OffsetPair> offsets() {
  return Collections.unmodifiableMap(_offsetPair);
}
```
Changed the definition to return an immutable map with the OffsetPair.

abhishekmendhekar:
After an offline discussion with @radai-rosenblatt , here is the conclusion. cc @smccauliff 

**API Behavior**
- `poll()` interface will only return any record processing exception
- on `seek*()` all the cached exceptions will be dropped
- if partition `P` is `pause()` and `P` had exception on last `poll()` then the next `poll()` will throw exception only for all unpaused partitions (excludes `P`). It'll also seek past the offset for all unpaused partitions that hit the exception (excludes `P`). Also, discard the cached exceptions.

**Additional optimizations**
- if `poll()` hit an exception when processing records and no other records have been fetched then `poll()` will throw an exception immediately instead of doing it in following `poll()` call.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -461,6 +481,48 @@ private long positionMain(TopicPartition partition, Duration timeout) {
...
+  private void seekToCurrentOffsetsOnRecordProcessingExceptions() {
+    // seek to offset which had an exception
+    if (_lastProcessedResult != null && _lastProcessedResult.hasException()) {
+      _lastProcessedResult.offsets().forEach((k, v) -> _kafkaConsumer.seek(k, v.getCurrentOffset()));
+    }
+  }
+
+  private void clearRecordProcessingException() {
+    if (_lastProcessedResult != null && _lastProcessedResult.hasException()) {
+      LOG.trace("Clearing all Record Processing Exceptions");
```
smccauliff: Can this be logged at warn or info level.  This would have been an exception delivered to a user.  Trace is almost as bad as just swallowing the exception.

smccauliff: And can you log the exceptions that are being dropped.

PR #125, opened 2019-06-26
Federated Clients: add a watchdog thread that polls for the creation of nonexistent topics in assignment/subscription and triggers re-assignment/subscription as necessary

jonlee2:


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,87 @@
...
+class FederatedSubscriptionState {
+  public enum SubscriptionType {
+    NONE, AUTO_TOPICS, AUTO_PATTERN, USER_ASSIGNED
+  }
+
+  private SubscriptionType _subscriptionType;
+
+  private Set<String> _subscription;
+
+  private Pattern _subscribedPattern;
```
radai-rosenblatt: do we want to also store the original string ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.kafka.common.TopicPartition;
+
+// Current subscription/assignment state at the federated level
+class FederatedSubscriptionState {
+  public enum SubscriptionType {
+    NONE, AUTO_TOPICS, AUTO_PATTERN, USER_ASSIGNED
```
radai-rosenblatt: should these be named closer to the kafka terminology: manual_assignment, subscription and wildcard_subscription ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").  See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.consumer;
```
radai-rosenblatt: kafka does not allow mixing multiple assignment types (i think?) yet this class does. would it be "safer" to protect this class from mixing multiple assignment types? even better - could this be made a parent interface with the distinct assignment types being concrete classes with final fields ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -40,6 +40,7 @@
   public static final String AUTO_COMMIT_INTERVAL_MS_CONFIG = ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG;
   public static final String AUTO_OFFSET_RESET_CONFIG = ConsumerConfig.AUTO_OFFSET_RESET_CONFIG;
   public static final String MAX_POLL_RECORDS_CONFIG = ConsumerConfig.MAX_POLL_RECORDS_CONFIG;
+  public static final String TOPIC_CREATION_POLL_INTERVAL_MS_CONFIG = "li.topic.creation.poll.interval.ms";
```
radai-rosenblatt: kafka clients already have a max.metadata.age or something - wouldnt it be better to default to that?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -92,6 +93,10 @@
 
   private static final String MAX_POLL_RECORDS_DOC = "The maximum number of records returned in a single call to poll().";
 
+  public static final String TOPIC_CREATION_POLL_INTERVAL_MS_DOC = "The frequency in milliseconds that a federated " +
```
radai-rosenblatt: name should be more general, because this will be (maybe) used to update metadata in general for stuff like wildcard subscriptions as well )which could also be affected by topic deletion)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -168,6 +173,11 @@
                 atLeast(1),
                 Importance.MEDIUM,
                 MAX_POLL_RECORDS_DOC)
+        .define(TOPIC_CREATION_POLL_INTERVAL_MS_CONFIG,
+                Type.INT,
+                300000,
```
radai-rosenblatt: as i said above - why does this need to be independant from max.metadata.age ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -113,6 +118,33 @@ public ClusterDescriptor getCluster() {
   // Number of config reload operations executed
   private volatile int _numConfigReloads;
 
+  private class PartitionsByClusterResult {
+    private Map<ClusterDescriptor, Collection<TopicPartition>> _clusterToPartitionsMap;
```
radai-rosenblatt: why at the partition level? why not just map cluster --> topic ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -260,27 +294,40 @@ synchronized public void assign(Collection<TopicPartition> partitions) {
...
+    Set<TopicPartition> partitionsSet = new HashSet<>(partitions);
+    _currentSubscription.setAssignment(partitionsSet);
+
+    // TODO: Throw an exception if there exists a partition whose topic exists but the specified partition number is
+    // out of range for that topic. Whether of not to throw an exception can be controlled by a property.
+    PartitionsByClusterResult partitionsByClusterResult = getPartitionsByCluster(partitionsSet);
+
+    _currentSubscription.setTopicsWaitingToBeCreated(partitionsByClusterResult.getNonexistentTopics());
+    if (!partitionsByClusterResult.getNonexistentTopics().isEmpty()) {
+      LOG.warn("the following partitions in the requested assignment currently do not exist: {}",
```
radai-rosenblatt: "topics" not "partitions"

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -771,4 +818,37 @@ private void ensureOpen() {
...
+    }
+
+    // Re-assign/subscribe
+    switch (_currentSubscription.getSubscriptionType()) {
+      case USER_ASSIGNED:
+        assign(_currentSubscription.getAssignment());
+        break;
+
+      default:
+        throw new IllegalStateException("Unsupportd subscription type");
```
radai-rosenblatt: you want to log here as this will throw on a thread out of the pool, which we may not see

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: These are what vanilla Kafka currently uses for its SubscriptionState.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: subscribe() for wildcard subscription directly takes Pattern, not a string.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
jonlee2: That makes sense. I'll use metadata.max.age.ms to control the poll interval as well.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
jonlee2: Will use metadata.max.age.ms.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
jonlee2: will use metadata.max.age.ms

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -771,4 +818,37 @@ private void ensureOpen() {
       throw new IllegalStateException("this federated consumer has already been closed");
     }
   }
+
+  private void maybeResubscribe() {
+    if (_closed || _currentSubscription.getSubscriptionType() == FederatedSubscriptionState.SubscriptionType.NONE ||
```
kehuum: Should we consider fallback mode here? If mario is not available, there is no need to continue the rest of the code. In the future we will have to change the registerFederatedClient API to return a boolean indicating whether it registers itself successfully within the timeout and use fallback mode if not. 

Can you please add a TODO here to check the register state?

jonlee2: Added a TODO

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
radai-rosenblatt: if those are the vanilla kafka terms, than stick to them.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,90 @@
...
+  public SubscriptionType getSubscriptionType() {
+    return _subscriptionType;
+  }
+
+  public void setTopicsWaitingToBeCreated(Set<String> topicsWaitingToBeCreated) {
+    _topicsWaitingToBeCreated = topicsWaitingToBeCreated;
+  }
+
+  public Set<String> getTopicsWaitingToBeCreated() {
+    return Collections.unmodifiableSet(new HashSet<>(_topicsWaitingToBeCreated));
```
radai-rosenblatt: no need to copy the underlying set ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.kafka.common.TopicPartition;
+
+// Current subscription/assignment state at the federated level
+public class FederatedSubscriptionState {
+  public enum SubscriptionType {
+    NONE, MANUAL_ASSIGNMENT, SUBSCRIPTION, PATTERN_SUBSCRIPTION
+  }
+
+  private SubscriptionType _subscriptionType;
```
radai-rosenblatt: no need to have the field - just hardcode the getter in the subclasses

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.kafka.common.TopicPartition;
+
+// Current subscription/assignment state at the federated level
+public class FederatedSubscriptionState {
```
radai-rosenblatt: this class shuold probably be abstract, with the constructors being protected

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -168,6 +172,12 @@
                 atLeast(1),
                 Importance.MEDIUM,
                 MAX_POLL_RECORDS_DOC)
+        .define(METADATA_MAX_AGE_CONFIG,
+                Type.INT,
+                5 * 60 * 1000,
```
radai-rosenblatt: is this available as a constant in kafka code somewhere ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -113,6 +119,33 @@ public ClusterDescriptor getCluster() {
   // Number of config reload operations executed
   private volatile int _numConfigReloads;
 
+  private class PartitionsByClusterResult {
```
radai-rosenblatt: some javadoc on what this is - the result of looking up the location of topics across multiple clusters in a group

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -146,6 +179,8 @@ private LiKafkaFederatedConsumerImpl(LiKafkaConsumerConfig configs, MetadataServ
 
     _mdsRequestTimeoutMs = configs.getInt(LiKafkaConsumerConfig.METADATA_SERVICE_REQUEST_TIMEOUT_MS_CONFIG);
     _maxPollRecordsForFederatedConsumer = configs.getInt(LiKafkaConsumerConfig.MAX_POLL_RECORDS_CONFIG);
+    _metadataMaxAgeMs = configs.getInt(LiKafkaConsumerConfig.METADATA_MAX_AGE_CONFIG);
+    _currentSubscription = new Unsubscribed();
```
radai-rosenblatt: nit - you probably want to only have a singleton static instance of Unsubscribed (also make sure its "topics not found" set is immutable)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -180,6 +215,18 @@ private LiKafkaFederatedConsumerImpl(LiKafkaConsumerConfig configs, MetadataServ
       }
       throw e;
     }
+
+    // Create a watchdog thread that polls the creation of nonexistent topics in the current assignment/subscription
+    // and re-assign/subscribe if any of them have been created since the last poll.
+    ScheduledExecutorService es = Executors.newSingleThreadScheduledExecutor();
```
radai-rosenblatt: thread names, daemonize them

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -260,27 +296,38 @@ synchronized public void assign(Collection<TopicPartition> partitions) {
...
+    // behavior.
     if (partitions.isEmpty()) {
       unsubscribe();
       return;
     }
 
-    Map<ClusterDescriptor, Collection<TopicPartition>> clusterToTopicPartitionsMap = getPartitionsByCluster(partitions);
+    // TODO: Throw an exception if there exists a partition whose topic exists but the specified partition number is
+    // out of range for that topic. Whether of not to throw an exception can be controlled by a property.
+    Set<TopicPartition> partitionsSet = new HashSet<>(partitions);
```
radai-rosenblatt: we need this copy because the method signature now expects a Set. this is a potentially significant mem copy - why not accept a collection ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -260,27 +296,38 @@ synchronized public void assign(Collection<TopicPartition> partitions) {
...
+    if (!partitionsByClusterResult.getNonexistentTopics().isEmpty()) {
+      LOG.warn("the following topics in the requested assignment currently do not exist: {}",
+          partitionsByClusterResult.getNonexistentTopics());
+    }
 
     // This assignment should replace the previous assignment. Since max.poll.records should be reset for consumers
     // for the new assignment, close all existing consumers first.
-    List<ClusterConsumerPair<K, V>> consumers = _consumers;
-    if (!consumers.isEmpty()) {
+    if (!_consumers.isEmpty()) {
```
radai-rosenblatt: in theory, if the set of clusters we talk to (hence the number of consumers) didnt change - we could avoide closing these, right ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -300,6 +347,8 @@ synchronized public void subscribe(Pattern pattern) {
 
   @Override
   synchronized public void unsubscribe() {
+    ensureOpen();
```
radai-rosenblatt: this is public, yet does not close/clear existing consumers (or pass the unsub() to them) - is this a potential bug if this is called externally ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -532,13 +581,17 @@ public void close() {
     close(CONSUMER_CLOSE_MAX_TIMEOUT_SECS, TimeUnit.SECONDS);
   }
 
-  void closeNoLock(long timeout, TimeUnit timeUnit) {
-    if (_closed) {
-      return;
-    }
+  @Override
+  synchronized public void close(Duration timeout) {
+    close(timeout.toMillis(), TimeUnit.MILLISECONDS);
```
radai-rosenblatt: i think we shuld do it the other way around - in upstream kafka the Duration-accepting methods have the better behaviour - we should have them be the "real" impls and have the 2-arg ones (long + unit) redirect to them.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -532,13 +581,17 @@ public void close() {
...
+  }
 
-    _closed = true;
+  // This is used for testing only
+  public FederatedSubscriptionState getCurrentSubscription() {
+    return _currentSubscription;
+  }
 
+  void closeNoLock(long timeout, TimeUnit timeUnit) {
```
radai-rosenblatt: private? (or explain why package protected)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -771,4 +818,40 @@ private void ensureOpen() {
       throw new IllegalStateException("this federated consumer has already been closed");
     }
   }
+
+  // TODO: check the client registration state and handle accordingly if it is the fallback mode
+  private void maybeResubscribe() {
```
radai-rosenblatt: names that start with maybe ..... can we call this "refreshSubscriptionMetadata" ? (or assignment metadata, or something along those lines)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -771,4 +818,40 @@ private void ensureOpen() {
...
+
+    boolean topicCreated = false;
+    for (String topic : _currentSubscription.getTopicsWaitingToBeCreated()) {
+      try {
+        if (_mdsClient.getClusterForTopic(topic, _clusterGroup, _mdsRequestTimeoutMs) != null) {
+          topicCreated = true;
+          break;
+        }
+      } catch (MetadataServiceClientException e) {
+        LOG.warn("cannot get cluster info for topic {}. ignoring...", topic);
```
radai-rosenblatt: include the exception

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -771,4 +818,40 @@ private void ensureOpen() {
...
+    }
+
+    // Re-assign/subscribe
+    switch (_currentSubscription.getSubscriptionType()) {
+      case MANUAL_ASSIGNMENT:
+        assign(((ManuallyAssigned) _currentSubscription).getAssignment());
+        break;
+
+      default:
+        LOG.error("unsupported subscription type: {}", _currentSubscription.getSubscriptionType().name());
```
radai-rosenblatt: nit - the last name() call isnt needed

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: Done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: Done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/FederatedSubscriptionState.java
jonlee2: Vanilla Kafka returns an unmodifiable copy, so I followed suit.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
jonlee2: Yes, in CommonClientConfigs, ProducerConfig, and ConsumerConfig. 

Actually, I didn't have to redefine them in LiKafkaConsumerConfig. As long as I define a default value here, it should be okay. Removed the definition and used CommonClientConfigs.METADATA_MAX_AGE_CONFIG/DOC here instead.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: Done. Updated LiKafkaConsumerImpl#close() which does the same thing as well.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: Added private and updated to use Duration.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: Done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: sure. done.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: done

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: Right. I am closing the existing consumers because the number of clusters to talk to for the new assignment may change.

A simpler and probably less bug-prone (but less ideal) approach would be to pre-create consumers for all clusters in the cluster group and fix max.poll.records based on that. Maybe we should do that? We probably don't need clusters groups with more than two clusters until NG takes over. What do you think?

This will require the new Mario query I asked before which will return all clusters in a cluster group.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: My bad. I didn't provide the full implementation of this method here, because I planned to do that when I add subscribe() implementation). And I added these lines just not to forget to change the subscription state.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
jonlee2: I spent a whole day debugging mock issues involving a mix of Collection and Set in my test code and decided to use a set for assignment set internally. I assumed assign() should not be a frequent operation and a copy should be ok?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
radai-rosenblatt: i hate the pre-creation idea. especially as the number of clusters may dynamically change

PR #128, opened 2019-07-12
Make KafkaConsumer poll(long) and poll(Duration) API visible to LiKaf…

alexwang08:
Currently `poll(long)` and `poll(Duration)` in `LiKafkaConsumerImpl` both call `KafkaConsumer.poll(Duration)`. We need to expose both of `KafkaConsumer`'s `poll()` in `LiKafkaConsumerImpl` so that `kafka-rest` can leverage them.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
@@ -234,8 +234,7 @@ public class LiKafkaConsumerImpl<K, V> implements LiKafkaConsumer<K, V> {
     _kafkaConsumer.unsubscribe();
   }
 
-  @Override
-  public ConsumerRecords<K, V> poll(long timeout) {
+  private ConsumerRecords<K, V> poll(long timeout, boolean includeMetadataInTimeout) {
     ConsumerRecords<K, V> processedRecords;
     // We will keep polling until timeout.
     long now = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -234,8 +234,7 @@ public void unsubscribe() {
     _kafkaConsumer.unsubscribe();
   }
 
-  @Override
-  public ConsumerRecords<K, V> poll(long timeout) {
+  private ConsumerRecords<K, V> poll(long timeout, boolean includeMetadataInTimeout) {
```
xiowu0: I was thinking making this API public. If we use this impl directly (e.g., kafka-rest), we can be more explicitly whether we should include the metatadata in timeout or not.  On the other head, this function is neither a public function in LikafkaConsumer interface nor in KafkaConsumer, it makes sense to be a private function.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -282,9 +285,14 @@ public void unsubscribe() {
     return processedRecords;
   }
 
+  @Override
```
xiowu0: deprecated this function ? since it replies on deprecated function?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
alexwang08: I'm inclined to keep the interface of `LiKafkaConsumerImpl` consistent and do not expose `poll(long, boolean)`

abhishekmendhekar: I guess if we want to keep the API pure, having this private method makes sense. This means we can avoid diverging from `KafkaConsumer` interface.

The bigger problem will be if and when `poll(long timeout)` is removed we'll be at the same junction as the old scala client. We need to document exactly why `poll(long timeout)` is needed second we need to push for a new API for `KafkaConsumer`

PR #130, opened 2019-07-19
Add support for config push at boot up time

kehuum:
Adds support for config push at boot up time. When federated client registers itself to Mario, Mario will return a set of configs it wants federated client to apply at boot up time. This set of configs is saved first then applied when actually creating per-cluster clients.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
@@ -116,11 +116,12 @@ public class LiKafkaFederatedConsumerImpl<K, V> implements LiKafkaConsumer<K, V>
 
   private int _nextClusterIndexToPoll;
 
-  // If client is in federated mode
-  private boolean _isFederatedMode;
+  // If client is in fallback mode. Fallback mode means this federated client is not able to talk to Conductor (due to
+  // either Conductor not deployed or network issue), so it will fallback to use the regular kafka client.
+  private boolean _isFallbackMode;
 
   // Number of consumers that successfully applied configs at boot up time, used for testing only
-  private int _numConsumersWithBootupConfigs = 0;
+  private Set<LiKafkaConsumer<K, V>> _numConsumersWithBootupConfigs = new HashSet<>();
 
   // The prefix of the client.id property to be used for individual consumers. Since the client id is part of the
   // consumer metric keys, we need to use cluster-specific client ids to differentiate metrics from different clusters.
@@ -227,11 +228,11 @@ public class LiKafkaFederatedConsumerImpl<K, V> implements LiKafkaConsumer<K, V>
       boolean registerSuccessful = _mdsClient.registerFederatedClient(this, _clusterGroup, configs.originals(), _mdsRequestTimeoutMs);
 
       if (registerSuccessful) {
-        _isFederatedMode = true;
+        _isFallbackMode = false;
         LOG.info("Register federated consumer group {} to Conductor succeeded, using federated mode", _clusterGroup.toString());
       } else {
-        _isFederatedMode = false;
-        LOG.info("Register federated consumer group {} to Conductor failed, using fallback mode", _clusterGroup.toString());
+        _isFallbackMode = true;
+        LOG.warn("Register federated consumer group {} to Conductor failed, using fallback mode", _clusterGroup.toString());
         // TODO: handle fallback mode here, possibly spin up regular LiKafkaConsumer here
       }
     } catch (Exception e) {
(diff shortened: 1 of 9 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -113,6 +116,12 @@ public ClusterDescriptor getCluster() {
 
   private int _nextClusterIndexToPoll;
 
+  // If client is in federated mode
+  private boolean _isFederatedMode;
```
radai-rosenblatt: as federated mode would be the common, expected, state, i'd name this after the less obvious possibility - isFallbackMode, with a javadoc explaining roughly what that means

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -113,6 +116,12 @@ public ClusterDescriptor getCluster() {
...
+
+  // Number of consumers that successfully applied configs at boot up time, used for testing only
+  private int _numConsumersWithBootupConfigs = 0;
```
radai-rosenblatt: make this some Set<> of those? might be more helpful in breakpoints

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -215,7 +224,16 @@ private LiKafkaFederatedConsumerImpl(LiKafkaConsumerConfig configs, MetadataServ
       // We assume that such information will be kept and used by the metadata service client itself.
       //
       // TODO: make sure this is not blocking indefinitely and also works when Mario is not available.
-      _mdsClient.registerFederatedClient(this, _clusterGroup, configs.originals(), _mdsRequestTimeoutMs);
+      boolean registerSuccessful = _mdsClient.registerFederatedClient(this, _clusterGroup, configs.originals(), _mdsRequestTimeoutMs);
```
radai-rosenblatt: i dislike the fact that success/failure is only communicated via a boolean here - can we distinguish between "cant find/access conductor (which would be a warning) and ssl handshake error here? maybe we should consider exceptions instead of a boolean here.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
```diff
@@ -215,7 +224,16 @@ private LiKafkaFederatedConsumerImpl(LiKafkaConsumerConfig configs, MetadataServ
...
+
+      if (registerSuccessful) {
+        _isFederatedMode = true;
+        LOG.info("Register federated consumer group {} to Conductor succeeded, using federated mode", _clusterGroup.toString());
+      } else {
+        _isFederatedMode = false;
+        LOG.info("Register federated consumer group {} to Conductor failed, using fallback mode", _clusterGroup.toString());
```
radai-rosenblatt: warn

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/metadataservice/MarioCommandCallbackImpl.java
```diff
@@ -31,19 +32,30 @@
...
+        if (clientType == LiKafkaFederatedClientType.FEDERATED_PRODUCER) {
           // Call producer reload config method
           ((LiKafkaFederatedProducerImpl) _federatedClient).reloadConfig(reloadConfigMsg.getConfigs(), reloadConfigMsg.getCommandId());
         } else {
           // call consumer reload config method
           ((LiKafkaFederatedConsumerImpl) _federatedClient).reloadConfig(reloadConfigMsg.getConfigs(), reloadConfigMsg.getCommandId());
         }
         break;
+      case REGISTER_RESPONSE:
+        // Upon receiving register response from conductor, federated client would save the configs from the message and apply the
```
radai-rosenblatt: corner case - this happens async, post construction, and the caller/constructor isnt blocked on this, right?
if so - is it possible that clients were created between construction and this point in time?

kehuum: Yes, the constructor is not blocked and the clients can be created in-between, therefore in applyBootupConfigFromConductor, it will check if clients are created, if yes, it will recreate them with the boot up configs.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaFederatedConsumerImpl.java
kehuum: I'm not sure how to use exceptions here since we are submitting the register job to an executor and check if the state has been changed to CONNECTED within the specified timeout. If we want to differentiate the different error cases, we need to instrument conductor client to throw different exceptions upon connecting to conductor server and all the way up to federated client.

PR #132, opened 2019-08-11
initial implementation of instrumented clients

radai-rosenblatt:
this PR adds "instrumented" kafka clients - these are decorators around kafka clients that maintain a connection to a MDS (conductor) and are capable of basic "remote control" - currently only config overrides.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
@@ -1,0 +1,789 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").  See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.consumer;
+
+import com.linkedin.kafka.clients.common.MetricsProxy;
+import com.linkedin.kafka.clients.utils.LiKafkaClientsUtils;
+import com.linkedin.mario.client.EventHandler;
+import com.linkedin.mario.client.SimpleClient;
+import com.linkedin.mario.client.SimpleClientState;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.Set;
+import java.util.StringJoiner;
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
+import org.apache.kafka.clients.consumer.OffsetCommitCallback;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * an instrumented consumer is a decorator around a single delegate client
+ * that supports registration with conductor, telemetry, and remote control
+ * (for example, pushing config changes from conductor side)
+ * @param <K>
+ * @param <V>
+ */
+public class LiKafkaInstrumentedConsumerImpl<K, V> implements Consumer<K, V>, EventHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(LiKafkaInstrumentedConsumerImpl.class);
+
+  private final long initialConnectionTimeoutMs = TimeUnit.SECONDS.toMillis(30);
+  private final ReadWriteLock delegateLock = new ReentrantReadWriteLock();
+  private final Properties baseConfig;
+  private final Function<Properties, Consumer<K, V>> consumerFactory;
+  private final String mdsUrl;
+  private final CountDownLatch initialConnectionLatch = new CountDownLatch(1);
+  private final MetricsProxy metricsProxy = new MetricsProxy() {
+    @Override
+    protected Map<MetricName, ? extends Metric> getMetrics() {
+      Consumer<K, V> delegate = LiKafkaInstrumentedConsumerImpl.this.delegate;
+      return delegate == null ? Collections.emptyMap() : delegate.metrics();
+    }
+  };
+  private volatile Long closedAt = null;
+  private volatile Map<String, String> configOverrides = null;
+  private volatile Consumer<K, V> delegate;
+  private final SimpleClient mdsClient;
+
+  private volatile Collection<String> subscribedTopics = null;
+  private volatile ConsumerRebalanceListener rebalanceListener = null;
+  private volatile Collection<TopicPartition> assignedPartitions = null;
+  private volatile Pattern subscriptionPattern = null;
+
+  public LiKafkaInstrumentedConsumerImpl(
+      Properties baseConfig,
+      Function<Properties, Consumer<K, V>> consumerFactory,
+      String mdsUrl
+  ) {
+    List<String> conversionIssues = new ArrayList<>(1);
+    this.baseConfig = baseConfig;
+    Map<String, String> translatedBaseConfig = LiKafkaClientsUtils.propertiesToStringMap(baseConfig, conversionIssues);
+    this.consumerFactory = consumerFactory;
+    this.mdsUrl = mdsUrl;
+
+    if (!conversionIssues.isEmpty()) {
+      StringJoiner csv = new StringJoiner(", ");
+      conversionIssues.forEach(csv::add);
+      LOG.error("issues translating consumer config to strings: {}", csv);
+    }
+
+    mdsClient = new SimpleClient(
+        mdsUrl,
+        TimeUnit.MINUTES.toMillis(1),
+        TimeUnit.HOURS.toMillis(1), translatedBaseConfig,
+        this
+    );
+
+    boolean tryFallback;
+    Exception issue = null;
+    try {
+      tryFallback = !initialConnectionLatch.await(initialConnectionTimeoutMs, TimeUnit.MILLISECONDS);
+    } catch (Exception e) {
+      tryFallback = true;
+      issue = e;
+    }
+
+    if (tryFallback) {
+      boolean delegateChanged = recreateDelegate(true);
+      if (delegateChanged) {
+        if (issue != null) {
+          LOG.error("exception waiting to contact {}, using user-provided configs as fallback", mdsUrl, issue);
+        } else {
+          LOG.error("unable to contact {} within timeout ({}), using user-provided configs as fallback", mdsUrl, initialConnectionTimeoutMs);
+        }
+      } else if (issue != null) {
+        //we got interrupted waiting, but apparently connection to mds was successful?
+        LOG.warn("exception waiting on MDS connection, yet connection succeeded?", issue);
+      }
+    }
+  }
+
+  @Override
+  public void stateChanged(SimpleClient client, SimpleClientState oldState, SimpleClientState newState) {
+    if (newState == SimpleClientState.REGISTERED) {
+      try {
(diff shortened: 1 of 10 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,583 @@
...
+   * there's also an assumption that unless closed, delegate is always != null
+   */
+  private void verifyOpen() {
+    if (isClosed()) {
+      //same exception thats thrown by a vanilla consumer
+      throw new IllegalStateException("This consumer has already been closed.");
+    }
+  }
+
+  private class MetricsProxy implements Map {
```
jonlee2: Why do we need MetricsProxy, instead of simply implementing metrics() as getDelegete() below?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,583 @@
...
+        }
+      } else if (issue != null) {
+        //we got interrupted waiting, but apparently connection to mds was successful?
+        LOG.warn("exception waiting on MDS connection, yet connection succeeded?", issue);
+      }
+    }
+  }
+
+  @Override
+  public void stateChanged(SimpleClient client, SimpleClientState oldState, SimpleClientState newState) {
```
jonlee2: Does this actually handle config reload after client registration is done? For federated clients, each method was defined as synchronized for thread-safety. That's not needed here?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -0,0 +1,342 @@
...
+   * there's also an assumption that unless closed, delegate is always != null
+   */
+  private void verifyOpen() {
+    if (isClosed()) {
+      //same exception thats thrown by a vanilla producer
+      throw new IllegalStateException("Cannot perform operation after producer has been closed");
+    }
+  }
+
+  private class MetricsProxy implements Map {
```
jonlee2: Can this be refactored? A TODO is fine.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -0,0 +1,342 @@
...
+      tryFallback = initialConnectionLatch.await(30, TimeUnit.SECONDS);
+    } catch (Exception e) {
+      tryFallback = true;
+      issue = e;
+    }
+
+    if (tryFallback) {
+      boolean delegateChanged = recreateDelegate(true);
+      if (delegateChanged) {
+        if (issue != null) {
```
navina: I don't understand this `if-else` block. What is the difference in the conveyed information in both log statements? 

I think it is useful to add a log message in the `catch` block near Line 83, along with the exception thrown.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
navina: Why do we need a MetricsProxy? It seems like the only usage until now, is to retrieve the metrics map, if the delegate exists.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -0,0 +1,342 @@
...
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.ProducerFencedException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class LiKafkaInstrumentedProducerImpl<K, V> implements Producer<K, V>, EventHandler {
```
navina: What is being "instrumented" with this producer implementation? some javadoc pls?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,583 @@
...
+        TimeUnit.MINUTES.toMillis(1),
+        TimeUnit.HOURS.toMillis(1),
+        translatedBaseConfig,
+        this
+    );
+
+    boolean tryFallback;
+    Exception issue = null;
+    try {
+      tryFallback = initialConnectionLatch.await(30, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
navina: > Shouldn't we try fallback if await() returns false (== timeout)? 

+1

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,583 @@
...
+    this.consumerFactory = consumerFactory;
+    this.mdsUrl = mdsUrl;
+
+    if (!conversionIssues.isEmpty()) {
+      StringJoiner csv = new StringJoiner(", ");
+      conversionIssues.forEach(csv::add);
+      LOG.error("issues translating consumer config to strings: {}", csv);
+    }
+
+    mdsClient = new SimpleClient(
```
navina: Suggestion: Is it worth adding a handle in `SimpleClient` to wait on connection to complete / timeout , instead of repeating this connection handling logic in both the producer and consumer implementations? 
Fyi, I haven't looked at `SimpleClient` . So, my comment may be incorrect.

MayureshGharat:
Hi @radai-rosenblatt , can you add description/motivation for this patch. It would help to understand what this PR is doing and would help in reviewing :)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/LiKafkaClientsUtils.java
```diff
@@ -101,4 +106,25 @@ public static void dumpStacksForAllLiveThreads(Set<Thread> threads) {
       }
     }
   }
+
+  public static Map<String, String> toStringMap(Properties props, List<String> errors) {
```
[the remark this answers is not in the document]
radai-rosenblatt: why? the art type is Properties

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/LiKafkaClientsUtils.java
```diff
@@ -101,4 +106,25 @@ public static void dumpStacksForAllLiveThreads(Set<Thread> threads) {
...
+      return null;
+    }
+    if (props.isEmpty()) {
+      return Collections.emptyMap();
+    }
+    Map<String, String> translated = new HashMap<>(props.size());
+    props.forEach((k, v) -> {
+      String sk = k == null ? null : k.toString();
+      String sv = v == null ? null : v.toString();
+      if (translated.containsKey(sk) && errors != null) {
```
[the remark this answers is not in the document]
radai-rosenblatt: your code is only equivalent if null values cannot exist (put can return null because null was the previous value under a key, not just because the key was not found).
however, after navina's comment above (Properties cannot have nulls at all) i can safely switch

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
radai-rosenblatt: its possible to write code that gets the metrics map and then holds onto that map for a looong time - longer than the delegate exists. using this proxy user code that maintains the "map" for long periods of time still gets the correct sensor values even if the delegate client has been recreated since

radai-rosenblatt: added javadocs to explain why the proxy exists. unclear on what you mean by "refactor"

jonlee2: You have the almost same implementation in LiKafkaInstrumentedConsumerImpl. I was wondering if this could/should be moved into some common class .

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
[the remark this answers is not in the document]
radai-rosenblatt: oops. thanks

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,583 @@
...
+          try {
+            oldConsumer.commitSync(Duration.ofSeconds(30));
+          } finally {
+            oldConsumer.close(Duration.ofSeconds(10));
+          }
+        } catch (Exception e) {
+          LOG.error("error closing old delegate consumer", e);
+        }
+      }
+      delegate = consumerFactory.apply(configOverrides);
```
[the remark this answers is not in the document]
radai-rosenblatt: construction of the delegate client is "outsourced", so not the instrumented client's problem

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
radai-rosenblatt: called by a single event handler thread from the mds client. implementation is thread safe. config push added via a different method

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
radai-rosenblatt: the exception caught is always printed, but either as error or as a warning (in case out timeout actually cared with a successful connection)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
radai-rosenblatt: i tried not to force users of SimpleClient into a "blocking constructor" pattern. its more flexible that way, at the cost of repeating the countdown logic (which isnt all that much, i think)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
navina: thanks!

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -0,0 +1,860 @@
...
+   * there's also an assumption that unless closed, delegate is always != null
+   */
+  private void verifyOpen() {
+    if (isClosed()) {
+      //same exception thats thrown by a vanilla consumer
+      throw new IllegalStateException("This consumer has already been closed.");
+    }
+  }
+
+//  private class MetricsProxy implements Map {
```
jonlee2: Remove this

radai-rosenblatt: oops

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
radai-rosenblatt: users can get a reference to the sensor map and hold onto it beyond the life of the delegate client it came from.

PR #135, opened 2019-08-22
Add kafka native record headers

viswamy:
This commit introduces two new kafka native headers
  1. "_t" -- which represents the timestamp at which the message was created by the producer.
  2. "_lm" -- which represents that this message is part of large message.


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2017 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").  See License in the project root for license information.
```
jonlee2: 2019

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  private PrimitiveEncoderDecoder() {
+
+  }
+
+  /**
+   * Encodes a long value into a {@link PrimitiveEncoderDecoder#LONG_SIZE} byte array
+   */
+  public static void encode(long value, byte[] bytes, int pos) {
+    if (bytes == null) {
```
jonlee2: You also want to check pos >= 0

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+   * Avoid instantiating PrimitiveEncoderDecoder class
+   */
+  private PrimitiveEncoderDecoder() {
+
+  }
+
+  /**
+   * Encodes a long value into a {@link PrimitiveEncoderDecoder#LONG_SIZE} byte array
+   */
+  public static void encode(long value, byte[] bytes, int pos) {
```
jonlee2: encodeLong?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+      throw new IllegalArgumentException("The input result cannot be null");
+    }
+
+    if (bytes.length < pos + LONG_SIZE) {
+      throw new IllegalArgumentException(
+          String.format("Not adequate bytes available to encode the long value(array length = %d, pos = %d", bytes.length, pos)
+      );
+    }
+
+    for (int i = 7; i >= 0; i--) {
```
jonlee2: use LONG_SIZE - 1 instead of 7

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+  public static byte[] encode(long value) {
+    byte[] data = new byte[LONG_SIZE];
+    encode(value, data, 0);
+    return data;
+  }
+
+  /**
+   * Encodes a int value into a {@link PrimitiveEncoderDecoder#INT_SIZE} byte array
+   */
+  public static void encode(int value, byte[] bytes, int pos) {
```
jonlee2: encodeInt?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+    if (bytes == null) {
+      throw new IllegalArgumentException("The input result cannot be null");
+    }
+
+    if (bytes.length < pos + INT_SIZE) {
+      throw new IllegalArgumentException(
+          String.format("Not adequate bytes available to encode the long value(array length = %d, pos = %d", bytes.length, pos)
```
jonlee2: s/long/int

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -238,6 +242,16 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
       Integer partition = producerRecord.partition();
       Future<RecordMetadata> future = null;
       UUID messageId = _uuidFactory.getUuid(producerRecord);
+
+      long eventTimestamp = System.currentTimeMillis();
```
jonlee2: _auditor below is using timestamp defined in line 241. Should it be used instead?

integration-tests/src/test/java/com/linkedin/kafka/clients/utils/LiKafkaClientsTestUtils.java
```diff
@@ -52,4 +56,15 @@ public static String getRandomString(int length) {
     }
     return stringBuiler.toString();
   }
+
+  public static Map<String, byte[]> convertHeaders(Headers headers) {
+    HashMap<String, byte[]> map = new HashMap<>();
+    for (Header header : headers) {
+      if (map.containsKey(header.key())) {
+        throw new IllegalStateException("Duplicate header found" + header.key());
```
jonlee2: Headers allow duplicate keys. As a generic util, I am not sure if we should throw an exception. What do you think?

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaProducerIntegrationTest.java
```diff
@@ -74,6 +78,12 @@ public void testSend() throws IOException, InterruptedException {
       while (messageCount < RECORD_COUNT && System.currentTimeMillis() < startMs + 30000) {
         ConsumerRecords<String, String> records = consumer.poll(100);
         for (ConsumerRecord<String, String> record : records) {
+          Map<String, byte[]> headers = LiKafkaClientsTestUtils.convertHeaders(record.headers());
+          assertTrue(headers.containsKey(Constants.TIMESTAMP));
+          assertTrue(
+              PrimitiveEncoderDecoder.decodeLong(headers.get(Constants.TIMESTAMP), 0)
+                  > System.currentTimeMillis() - TimeUnit.DAYS.toMillis(1)
```
jonlee2: Can we record start and end timestamps of producer.send() above and make sure the timestamp in the header is within that range?

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -116,6 +115,20 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
         consumptionStarted = true;
       }
       for (ConsumerRecord<String, String> consumerRecord : records) {
+        // Verify headers
+        Map<String, byte[]> headers = LiKafkaClientsTestUtils.convertHeaders(consumerRecord.headers());
+        assertTrue(headers.containsKey(Constants.TIMESTAMP));
+        assertEquals(8, headers.get(Constants.TIMESTAMP).length);
+        assertTrue(
+            PrimitiveEncoderDecoder.decodeLong(headers.get(Constants.TIMESTAMP), 0)
+                > System.currentTimeMillis() - TimeUnit.DAYS.toMillis(1)
```
jonlee2: Can we record start and end timestamps of producer.send() above and make sure the timestamp in the header is within that range?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -417,6 +423,24 @@ public void close(long timeout, TimeUnit timeUnit) {
...
+        Header header = iterable.next();
+        largeMessageHeaderValue = LargeMessageHeaderValue.fromBytes(header.value());
+      }
+      // Once the large message is parsed, remove any such key from record headers
+      headers.remove(Constants.LARGE_MESSAGE);
+      if (largeMessageHeaderValue != null) {
+        largeMessageHeaderValue = new LargeMessageHeaderValue(
+            LargeMessageHeaderValue.Type.LM_ENCRYPTION_AVRO,
+            largeMessageHeaderValue.getUuid(),
+            -1,
```
jonlee2: What is the reason for wiping out the sequence id?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -417,6 +423,24 @@ public void close(long timeout, TimeUnit timeUnit) {
...
       _partitionConsumerHighWatermark.computeIfAbsent(tp, _storedConsumerHighWatermark)._currentConsumerHighWatermark = consumerRecord.offset();
+      LargeMessageHeaderValue largeMessageHeaderValue = null;
+      Headers headers = new RecordHeaders(consumerRecord.headers());
+      Iterator<Header> iterable = headers.headers(Constants.LARGE_MESSAGE).iterator();
+      if (iterable.hasNext()) {
+        Header header = iterable.next();
+        largeMessageHeaderValue = LargeMessageHeaderValue.fromBytes(header.value());
+      }
+      // Once the large message is parsed, remove any such key from record headers
+      headers.remove(Constants.LARGE_MESSAGE);
```
jonlee2: I think lines 434-443 can actually be put inside the if block in line 429-432?

.gitignore
```diff
@@ -1,6 +1,7 @@
 #gradle
 .gradle/
 build/
+integration-tests/.mario/
```
radai-rosenblatt: thank you

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -28,8 +31,8 @@
 import java.util.Properties;
 import java.util.Set;
 
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.*;
```
radai-rosenblatt: do we like star imports ?

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
radai-rosenblatt: +1, better bound than 1 day

integration-tests/src/test/java/com/linkedin/kafka/clients/utils/LiKafkaClientsTestUtils.java
radai-rosenblatt: i agree. we should just not allow any dup "_*" headers

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
viswamy: Nope. This is just intellij messing with me.

navina:
@viswamy What is the motivation for this change ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
viswamy: At first, I was under the impression that the timestamps should be completely dictated by the client. This is to avoid end users abusing the header value. Certain downstream systems like auditing may be impacted if timestamps are way off (or invalid). But on second thought, there is a very valid use case in change capture systems who are more interested in the timestamp at which a particular object was updated in the backend store as opposed to the current timestamp. I will just make it equivalent to TmeAuditor's timestamp and any -ve effects that this may have will be handled by the downstream.

integration-tests/src/test/java/com/linkedin/kafka/clients/utils/LiKafkaClientsTestUtils.java
viswamy: This is test util -- so it won't be imported into the final jar(IMO). 

I have changed the util to only consider the special headers with "_" prefix.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
viswamy: Because the large message will not have any sequence number associated with it. It is only the internal splits that will have segment ids starting from zero. This is the record that will be returned to the end-user.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
viswamy: I have simplified the code. What do you think now?

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -116,6 +117,18 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
         consumptionStarted = true;
       }
       for (ConsumerRecord<String, String> consumerRecord : records) {
+        // Verify headers
+        Map<String, byte[]> headers = LiKafkaClientsTestUtils.fetchSpecialHeaders(consumerRecord.headers());
+        assertTrue(headers.containsKey(Constants.TIMESTAMP));
+        assertEquals(8, headers.get(Constants.TIMESTAMP).length);
```
jonlee2: use PrimitiveEncoderDecoder.LONG_SIZE instead of 8?

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -116,6 +117,18 @@ public void onCompletion(RecordMetadata recordMetadata, Exception e) {
...
+        long eventTimestamp = PrimitiveEncoderDecoder.decodeLong(headers.get(Constants.TIMESTAMP), 0);
+        assertTrue(eventTimestamp >= startTime && eventTimestamp <= System.currentTimeMillis());
```
jonlee2: Why not get System.currentTimeMillis() before the for loop and use it?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,125 @@
...
+    byte[] data = new byte[INT_SIZE];
+    encodeInt(value, data, 0);
+    return data;
+  }
+
+  public static int decodeInt(byte[] bytes, int pos) {
+    if (bytes == null) {
+      throw new IllegalArgumentException("bytes cannot be null");
+    }
+
```
jonlee2: Add pos check to be consistent with others

integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
viswamy: IMO, it just pollutes the test case with more variables.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
viswamy: Good catch.

.gitignore
```diff
@@ -1,6 +1,7 @@
 #gradle
 .gradle/
 build/
+*.mario/
```
radai-rosenblatt: * not needed - just ".mario/" (without the quotes) should work i think

integration-tests/src/test/java/com/linkedin/kafka/clients/utils/LiKafkaClientsTestUtils.java
```diff
@@ -52,4 +56,26 @@ public static String getRandomString(int length) {
...
+    HashMap<String, byte[]> map = new HashMap<>();
+    for (Header header : headers) {
+
+      if (!header.key().startsWith("_")) {
+        // skip any non special header
+        continue;
+      }
+
+      if (map.containsKey(header.key())) {
+        throw new IllegalStateException("Duplicate special header found" + header.key());
```
radai-rosenblatt: nit - space

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private static Type typeOf(byte value) {
+    switch (value) {
+      case (byte) 0: return Type.LEGACY;
+      default: throw new IllegalArgumentException("Unknown type found! = " + value);
+    }
+  }
+
+  public enum Type {
```
radai-rosenblatt: we want this to be pluggable - meaning we'd like to be able to register new types (and their encoders/decoders) from non open-source code. i'd suggest not using an enum for this reason (could still have 0 as a const)

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
```diff
@@ -50,12 +54,13 @@ public MessageSplitterImpl(int maxSegmentSize,
 
   @Override
   public List<ProducerRecord<byte[], byte[]>> split(String topic, Integer partition, UUID messageId, byte[] key, byte[] serializedRecord) {
-    return split(topic, partition, null, messageId, key, serializedRecord);
+    return split(topic, partition, null, messageId, key, serializedRecord, null);
```
radai-rosenblatt: i'd declare this signature deprecated

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/Constants.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").  See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.utils;
+
+public class Constants {
+  // The variables reserved by kafka for auditing purpose
+  public static final String TIMESTAMP = "_t";
```
radai-rosenblatt: i'd call them TIMESTAMP_HEADER, LARGE_MSG_HEADER etc

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,133 @@
...
+   * Avoid instantiating PrimitiveEncoderDecoder class
+   */
+  private PrimitiveEncoderDecoder() {
+
+  }
+
+  /**
+   * Encodes a long value into a {@link PrimitiveEncoderDecoder#LONG_SIZE} byte array
+   */
+  public static void encodeLong(long value, byte[] bytes, int pos) {
```
radai-rosenblatt: i'd call the "bytes" parameter "output"

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
```diff
@@ -0,0 +1,104 @@
...
+    // This represents that both large message segments and encryption use avro serialization/deserialization
```
navina: the encryption/decryption interface is purely exposed to linkedin specific client. So, the comment and the implied meaning of this byte seems out of place. 

What is the "type" field in the header trying to represent? Is it the serialization format being used for large message ? Or is it used to represent if the payload is encrypted or not?  It can't imply both. 
I would suggest using bit-masks, if necessary to indicate different flags/indicators.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
navina: Wouldn't the same logic apply for `getNumberOfSegments` ? I mean it just feels like we are exposing the internal details unintentionally, if we are exposing the  idea of segments.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -0,0 +1,133 @@
...
+ * Copyright 2019 LinkedIn Corp. Licensed under the BSD 2-Clause License (the "License").
 See License in the project root for license information.
+ */
+
+package com.linkedin.kafka.clients.utils;
+
+/**
+ * Class to encode/decode primitive integer/long types.
+ */
+public class PrimitiveEncoderDecoder {
```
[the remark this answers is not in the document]
viswamy: 1) platform-independent -- we can port this logic to any language we want(in future).
2) No additional object allocations. If we were to go with jdk based, one way is to allocate bytebuffers and use #putLong, #putInt etc,. but it ends up creating multiple intermediate objects and causes GC pressure. The class was designed specifically to address that concern.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -238,6 +242,15 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
       Integer partition = producerRecord.partition();
       Future<RecordMetadata> future = null;
       UUID messageId = _uuidFactory.getUuid(producerRecord);
+
+      Headers headers = producerRecord.headers();
+      if (headers == null) {
+        headers = new RecordHeaders();
+      }
+      // Remove any header that maybe using the key for audit event timestamp or large message.
```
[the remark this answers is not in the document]
viswamy: "_" prefixed headers are meant to be used internally (yet to update the documentation). That said, we don't really want end-users to set these values. They can communicate the record timestamp via ProducerRecord#timestmap() though. // Sticking with the earlier auditing behavior.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
viswamy: The number of segments is required for the auditing use case(i.e,. count the actual number of messages by using a Likafka client). A case can be made that UUID isn't required but since we are using a certain header for large message -- might as well use the same.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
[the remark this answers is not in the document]
viswamy: > Also, should the actual timestamp always be the System instead of ProducerRecord.timestamp()?

This is what I had in my earlier implementation but changed it so that it is now equivalent to the Auditor behavior.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
viswamy: I have removed the Type and made it a variable. We can add more variables with specific values as and when we implement the features.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
navina: >  A case can be made that UUID isn't required but since we are using a certain header for large message -- might as well use the same. 
I am not following the line of logic here. I would rather not expose anything internal as much as possible. That's why I am concerned about exposing the number of segments. It seems to serve the purpose for auditing. I don't see any use-case for exposing uuid.

viswamy: Set the UUID to an empty value.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
```diff
@@ -0,0 +1,87 @@
...
+        largeMessageHeaderValue.getNumberOfSegments(),
+        serialized,
+        1 + 2 * PrimitiveEncoderDecoder.LONG_SIZE + PrimitiveEncoderDecoder.INT_SIZE
+    );
+    return serialized;
+  }
+
+  public static LargeMessageHeaderValue fromBytes(byte[] bytes) {
+    byte type = bytes[0];
+    long leastSignificantBits = PrimitiveEncoderDecoder.decodeLong(bytes, 1);
```
abhishekmendhekar: nit: using a variable to keep track of the current offset in the byte array makes it easier to read the code.

```
int bytesOffset = 0;
bytesOffset += 1; // for lsb of UUID
bytesOffset += PrimitiveEncoderDecoder.LONG_SIZE; // for msb of UUID
bytesOffset += PrimitiveEncoderDecoder.LONG_SIZE; // for segment number
bytesOffset += PrimitiveEncoderDecoder.INT_SIZE; // for number of segments
```

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -417,6 +422,21 @@ public void close(long timeout, TimeUnit timeUnit) {
...
+      Headers headers = new RecordHeaders(consumerRecord.headers());
+      Header largeMessageHeader = headers.lastHeader(Constants.LARGE_MESSAGE_HEADER);
+      if (largeMessageHeader != null) {
+        LargeMessageHeaderValue largeMessageHeaderValue = LargeMessageHeaderValue.fromBytes(largeMessageHeader.value());
+        // Once the large message header value is parsed, remove any such key from record headers
+        headers.remove(Constants.LARGE_MESSAGE_HEADER);
+        largeMessageHeaderValue = new LargeMessageHeaderValue(
+            largeMessageHeaderValue.getType(),
+            LargeMessageHeaderValue.EMPTY_UUID,
+            -1,
```
abhishekmendhekar: nit: Can we make -1 a constant? Makes it easier to understand what `-1` means.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/MessageSplitterImpl.java
```diff
@@ -96,8 +108,26 @@ public MessageSplitterImpl(int maxSegmentSize,
...
+
       ProducerRecord<byte[], byte[]> segmentProducerRecord =
-          new ProducerRecord<>(topic, partition, timestamp, segmentKey, segmentValue);
+          new ProducerRecord<>(
+              topic,
+              partition,
+              timestamp,
+              segmentKey,
+              segmentValue,
+              temporaryHeaders.add(Constants.LARGE_MESSAGE_HEADER, LargeMessageHeaderValue.toBytes(largeMessageHeaderValue))
```
abhishekmendhekar: nit: I suggest calling `add()` followed by `remove()` making it easier to understand that that the large message header is being overwritten.

PR #140, opened 2019-09-06
Introduce safe offset record header

viswamy:
What and why?
For every record consumed from li-apache-kafka-client consumer, the record will be annotated with a safe offset header such that systems like brooklin aren't forced to use LiKafkaConsumer interface (which has #safeOffsets() api) but instead can just work with Consumer interface and the headers.


viswamy:
Will add an integ test to cover this logic. Any early feedback is welcome!

PR #156, opened 2019-11-07
make instrumented consumer behave the same as vanilla consumer around concurrent poll and close

radai-rosenblatt:
note that vanilla clients would throw an exception on the thread calling close() in this case.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
@@ -5,6 +5,8 @@
 package com.linkedin.kafka.clients.consumer;
 
 import com.linkedin.kafka.clients.common.MetricsProxy;
+import com.linkedin.kafka.clients.utils.CloseableLock;
+import com.linkedin.kafka.clients.utils.KafkaConsumerLock;
 import com.linkedin.kafka.clients.utils.LiKafkaClientsUtils;
 import com.linkedin.mario.client.EventHandler;
 import com.linkedin.mario.client.SimpleClient;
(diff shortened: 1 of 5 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -776,6 +744,9 @@ private boolean proceedClosing() {
     if (isClosed()) {
       return false;
     }
+    closing = true;
```
ambroff: I'm not entirely sure this is correct, especially since proceedClosing() can be called from Close() which doesn't acquire a lock. I think it would be safest to replace this `volatile boolean` with an AtomicBoolean and use [AtomicBoolean#compareAndSet(boolean, boolean)](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicBoolean.html#compareAndSet(boolean,%20boolean)) to guarantee that you don't have a situation where two threads race to set `closing = true`.

ambroff: I at least think it would simplify the logic in proceedClosing a bit

ambroff: Just as an example problem, you could have two threads setting close, assigning closer, and then try to acruire delegateLock. Only one will get to that point, but the code that acquired delegateLock first *might not* be the last one to set `closer`.

So I think this is better:

```
if (!closing.compareAndSet(false, true)) {
  return;
}
```

Or however you use that correctly.

radai-rosenblatt: i've reworked the patch, and addressed these issues (i think)

PR #163, opened 2020-01-12
check for null before closing delegates of instrumented clients

radai-rosenblatt:


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -723,9 +726,12 @@ public void close(long timeout, TimeUnit unit) {
         return;
       }
       try {
-        delegate.close(timeout, unit);
+        Consumer<K, V> delegate = this.delegate;
+        if (delegate != null) {
+          delegate.close(timeout, unit);
```
Lincong: Can we add some logging here to show what the time out is and how much time does it actually take for the delegate consumer to shut down?
```java
          LOG.info("Trying to close the consumer delegate in {} {}", timeout, unit);
          long closeStartTimeMs = System.currentTimeMillis();
          delegate.close(timeout, unit);
          LOG.info("Close the consumer delegate finished. Took {} ms", System.currentTimeMillis() - closeStartTimeMs);

```

radai-rosenblatt: info is too verbose for this. users who care about timeouts can set them and be "notified" by an exception. otherwise this is profiling information

radai-rosenblatt: as for doClose. there are actually subtle differences upstream between the various close() invocations (the one that accept Duration honor more timeouts), so choosing a single upstream path actually changes behaviour

PR #171, opened 2020-04-21
use extended serde interface for sdc testing

sudoa:
Compression/Decompression will be implemented by extending Extended{Serializer, Deserializer} interface and adding a delegate avro serializer/deserializer. So Serde with (de)compression will be passed in from "Linkedin bits" and used in the client. At produce time, we first serialize and then compress, at consume time, we first decompress then serialize.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
@@ -25,7 +25,6 @@ import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.ExtendedDeserializer;
-import org.apache.kafka.common.serialization.ExtendedSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -421,7 +423,12 @@ public long getRecordsSkipped() {
     K key = _keyDeserializer.deserialize(tp.topic(), consumerRecord.key());
     byte[] valueBytes = parseAndMaybeTrackRecord(tp, consumerRecord.offset(), consumerRecord.value());
     if (valueBytes != INCOMPLETE_RESULT) {
-      V value = _valueDeserializer.deserialize(tp.topic(), valueBytes);
+      V value;
```
radai-rosenblatt, line 426: perf nit - instead of performing an instanceof per record, can we do this once and use a delegate here to call the "right" method?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -290,7 +291,12 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
       byte[] serializedValue;
       byte[] serializedKey;
       try {
-        serializedValue = _valueSerializer.serialize(topic, value);
+        // ExtendedSerializer is used to test Share dictionary compression
+        if (_valueSerializer instanceof ExtendedSerializer) {
```
radai-rosenblatt, line 295: same as above comemnt

PR #172, opened 2020-05-01
Adding original message size to the large message header value

navina:
`LargeMessageHeaderValue` is missing `messageSizeInBytes`, which indicates the size of the original message. This is useful to validate the segments during assembling on the consumer side. <br/>
With the existing code, we will see failures when we try to assemble based on the metadata from the large message header.  Not sure why this left out in the original patch that introduced the header value. Adding it now to ensure that segment validation can be done.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
@@ -10,10 +10,10 @@ import java.util.UUID;
 
 /**
  * This class represents the header value for a large message.
- * Every large message header takes up 25 bytes and is structured as follows
+ * Every large message header takes up 29 bytes and is structured as follows
  *
- * | Type   | UUID     | segmentNumber | numberOfSegments |
- * | 1 byte | 16 bytes | 4 bytes       | 4 bytes          |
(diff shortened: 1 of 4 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/common/LargeMessageHeaderValue.java
```diff
@@ -81,6 +95,11 @@ public static LargeMessageHeaderValue fromBytes(byte[] bytes) {
     int segmentNumber = PrimitiveEncoderDecoder.decodeInt(bytes, byteOffset);
     byteOffset += PrimitiveEncoderDecoder.INT_SIZE;
     int numberOfSegments = PrimitiveEncoderDecoder.decodeInt(bytes, byteOffset);
-    return new LargeMessageHeaderValue(type, new UUID(mostSignificantBits, leastSignificantBits), segmentNumber, numberOfSegments);
+    if (bytes.length == 29) {
```
kehuum, line 98: nit: Can we make this 29 a macro or something? Later when we debug or need to revisit the code, I might forgot what this 29 is.

navina: sg! will fix it.

PR #173, opened 2020-05-11
Change alert level of mario client connection timeout from error to warning for LiKafkaInstrumentedConsumer

zsong-li:
Change alert level of mario client connection timeout from error to warning for LiKafkaInstrumentedConsumer

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
@@ -150,7 +150,7 @@ public class LiKafkaInstrumentedConsumerImpl<K, V> implements DelegatingConsumer
           LOG.error("exception waiting to contact {}, using user-provided configs as fallback",
               mdsClient.getLastAttemptedMarioUrl(), issue);
         } else {
-          LOG.error("unable to contact {} within timeout ({}), using user-provided configs as fallback",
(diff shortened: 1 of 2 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaInstrumentedConsumerImpl.java
```diff
@@ -150,7 +150,7 @@ public LiKafkaInstrumentedConsumerImpl(
           LOG.error("exception waiting to contact {}, using user-provided configs as fallback",
               mdsClient.getLastAttemptedMarioUrl(), issue);
         } else {
-          LOG.error("unable to contact {} within timeout ({}), using user-provided configs as fallback",
+          LOG.warn("unable to contact {} within timeout ({}), using user-provided configs as fallback",
```
jonlee2, line 153: Why do you change the severity for consumer only?

zsong-li: We have several tickets reported the instability of conductor service and hereby affected the mario client connection for this consumer. As @sudoa mentioned above, this is a temp solution now but probably we can consider to extend it to long-term as well since even if the mario failed to talk to the conductor service, the produce/consume should not be affected.

jonlee2: I mean there's similar code in LiKafkaInstrumentedProducerImpl.java. Why do you not change it as well?

PR #178, opened 2020-06-29
Added a new consumer metric - consumerOffsetWatermarkSpan to represent "distance between the safe offset and watermark"

navina:
In order to identify a partition "stall" due to partial large message assembly (and not due to other problems in the consumer), we are introducing a metric called "consumer offset watermark span".  For each partition, the span distance is computed as (`watermark minus safeOffset`). Consumer offset watermark span is defined as the sum of all span distance for assigned partitions that have currently undelivered large messages in the consumer. 
When the consumer is progressing with no partial large messages in buffer, its span distance should be 0 because watermark is same as safeOffset.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
@@ -148,22 +148,22 @@ public class LiKafkaConsumerImpl<K, V> implements LiKafkaConsumer<K, V> {
       }
     };
 
-    MetricName consumerSnagMetricName = new MetricName(
-        "consumer-snag",
+    MetricName consumerOffsetWatermarkSpan = new MetricName(
+        "consumer-offset-watermark-span",
         "lnkd",
-        "sum of snag distance values for all partitions, where snag distance of a partition "
+        "sum of offset-watermark-span for all partitions, where offset-watermark-span of a partition "
             + "shows how far behind the safe offset of a partition is from its watermark.",
         Collections.singletonMap("client-id", _clientId)
     );
-    Metric consumerSnagMetric = new Metric() {
+    Metric consumerOffsetWatermarkSpanMetric = new Metric() {
       @Override
       public MetricName metricName() {
-        return consumerSnagMetricName;
+        return consumerOffsetWatermarkSpan;
       }
 
       @Override
       public double value() {
-        return (double) _consumerRecordsProcessor.getConsumerSnag();
+        return (double) _consumerRecordsProcessor.getConsumerOffsetWatermarkSpan();
       }
 
       @Override
@@ -173,7 +173,7 @@ public class LiKafkaConsumerImpl<K, V> implements LiKafkaConsumer<K, V> {
     };
 
     _extraMetrics.put(skippedRecordsMetricName, skippedRecordsMetric);
-    _extraMetrics.put(consumerSnagMetricName, consumerSnagMetric);
+    _extraMetrics.put(consumerOffsetWatermarkSpan, consumerOffsetWatermarkSpanMetric);
 
     try {
 
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
@@ -79,7 +79,7 @@ public class LiKafkaConsumerImpl<K, V> implements LiKafkaConsumer<K, V> {
   private final long _autoCommitInterval;
   private final LiOffsetResetStrategy _offsetResetStrategy;
   private long _lastAutoCommitMs;
-  private final Map<MetricName, Metric> _extraMetrics = new HashMap<>(1);
+  private final Map<MetricName, Metric> _extraMetrics = new HashMap<>(2);
 
   private ConsumerRecordsProcessResult<K, V> _lastProcessedResult;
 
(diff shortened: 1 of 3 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/DeliveredMessageOffsetTracker.java
```diff
@@ -192,11 +202,34 @@ private boolean hasDeliveredMessages(PartitionOffsetTracker offsetTracker) {
     }
 
     void updateCurrentSafeOffset(long currentSafeOffset) {
+      if (_delivered != -1) {
+        if (currentSafeOffset == _delivered) {
+          _partitionSnag = 0;
+        } else {
+          if (currentSafeOffset > _currentSafeOffset) { // safe offset moves forward
+            _partitionSnag -= (currentSafeOffset - _currentSafeOffset);
```
radai-rosenblatt, line 210: cant we always say that snag = _delivered - _safeOffset  and just calculate it instead of trying to "diff" it?

navina: Sure we can do that. I think the complexity of the "Diff" is not needed for this coarse metric.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -148,7 +148,32 @@ public Object metricValue() {
       }
     };
 
+    MetricName consumerOffsetWatermarkSpan = new MetricName(
+        "consumer-offset-watermark-span",
+        "lnkd",
+        "sum of offset-watermark-span for all partitions, where offset-watermark-span of a partition "
+            + "shows how far behind the safe offset of a partition is from its watermark.",
```
radai-rosenblatt, line 155: nit - missing a space?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -87,6 +87,15 @@ boolean isCommittedOffsetDefined() {
...
   private long recordsSkipped = 0;
+  /**
+   * Gives an idea about how far behind the safe offset of a partition is from the watermark. Hence, the
+   * offset-watermark span or a partition is computed as (watermark minus safeoffset).
+   * This metric is a sum of
+   * {@link com.linkedin.kafka.clients.largemessage.DeliveredMessageOffsetTracker#offsetWatermarkSpan} for all
+   * assigned partitions that have currently undelivered large messages in the consumer. It is computed on query of
+   * the metric each time and hence, stores the last known value for this metric.
+   */
+  private long _consumerOffsetWatermarkSpan = -1;
```
radai-rosenblatt, line 98: nit - wouldnt 0 be a safer default?

smccauliff, line 98: Metrics are often accessed from different threads, even in the consumer, and so shouldn't this be volatile.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -433,6 +442,15 @@ public long getRecordsSkipped() {
     return recordsSkipped;
   }
 
+  public long getConsumerOffsetWatermarkSpan() {
+    long span = 0;
+    for (TopicPartition tp : knownPartitions()) {
```
smccauliff, line 447: This is probably also not safe for multithreaded access either since the partition assignment can change while this is running.

radai-rosenblatt: true, but these are sensor readings, they can mess up. i think some kafka sensors already do. also the value of knownPartitions() may not hold for the entire iteration

smccauliff: Yes, and I've had to fix ConcurrentModificationException stack traces many times as they will occasionally fill users' logs.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
navina: Probably.  I think this local variable is not required as the metric is computed in each call to `getConsumerOffsetWatermarkSpan`

navina: Makes sense. Looking at the code, though, this variable doesn't need to exist since the value is computed each time in the `getConsumerOffsetWatermarkSpan` method.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
navina: @smccauliff Are you saying the access to `knownPartitions()` is not thread safe? 

@radai-rosenblatt It should be fine if the assigned partitions don't hold for the entire iteration. The value returned by `deliveryMessageOffsetTracker` is 0, if a topic partition is not being tracked.

smccauliff: This is what I see:
```java
  public Set<TopicPartition> knownPartitions() {
    return Collections.unmodifiableSet(_partitionConsumerHighWatermark.keySet());
  }
```
If _partitionConsumerHighWatermark is modified during the computation, by the consumer thread, then the metrics reporting thread can get ConcurrentModificationException.  I've seen similar problems before.  Either the code needs to be made thread safe or ConcurrentModificationException needs to be caught and then the computation needs to be retried.

navina: I made the change with a memcpy instead of handling `ConcurrentModificationException`. Let me know if that is ok.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/largemessage/ConsumerRecordsProcessor.java
```diff
@@ -433,6 +434,24 @@ public long getRecordsSkipped() {
...
+   * offset-watermark span or a partition is computed as (watermark minus safeoffset).
+   * This metric is a sum of
+   * {@link com.linkedin.kafka.clients.largemessage.DeliveredMessageOffsetTracker#offsetWatermarkSpan} for all
+   * assigned partitions that have currently undelivered large messages in the consumer. It is computed on query of
+   * the metric each time.
+   */
+  public long getConsumerOffsetWatermarkSpan() {
+    long span = 0;
+    // mem copy to avoid ConcurrentModificationException, in case known partition set changes
+    final Set<TopicPartition> knownPartitions = new HashSet<>(knownPartitions());
```
smccauliff, line 448: The HashSet constructor calls addAll which does this:
```java
 public boolean addAll(Collection<? extends E> c) {
        boolean modified = false;
        for (E e : c)
            if (add(e))
                modified = true;
        return modified;
    }
```

This can still get CME because HashSet<>(Collection<>) will use an iterator to make a copy.   There needs to be some kind of synchronization in order to avoid this or use some kind of concurrent data structure.

PR #182, opened 2020-07-31
Add exception.on.invalid.offset config to allow throwing exception when auto.offset.reset is set to some policy

kehuum:
Add an exception.on.invalid.offset config to allow throw the InvalidOffsetException even when auto.offset.reset for LiKafkaConsumer is set to some policy (earliest/latest/liclosest). This is to allow users to handle this exception while let LiKafkaConsumer deal with the internal state of large message.

This is just a temporary solution and will be discarded when large message support has been refactored

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
@@ -33,6 +33,7 @@ public class LiKafkaConsumerConfig extends AbstractConfig {
   public static final String MESSAGE_ASSEMBLER_EXPIRATION_OFFSET_GAP_CONFIG = "message.assembler.expiration.offset.gap";
   public static final String MAX_TRACKED_MESSAGES_PER_PARTITION_CONFIG = "max.tracked.messages.per.partition";
   public static final String EXCEPTION_ON_MESSAGE_DROPPED_CONFIG = "exception.on.message.dropped";
+  public static final String EXCEPTION_ON_INVALID_OFFSET_RESET_CONFIG = "exception.on.invalid.offset.reset";
   public static final String TREAT_BAD_SEGMENTS_AS_PAYLOAD_CONFIG = "treat.bad.segments.as.payload";
   public static final String KEY_DESERIALIZER_CLASS_CONFIG = ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
   public static final String VALUE_DESERIALIZER_CLASS_CONFIG = ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG;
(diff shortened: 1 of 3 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -319,6 +322,9 @@ public void unsubscribe() {
         }
       } catch (OffsetOutOfRangeException | NoOffsetForPartitionException oe) {
         handleInvalidOffsetException(oe);
+        if (_throwExceptionOnInvalidOffsets) {
```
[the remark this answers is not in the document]
kehuum: No, the behavior we want is to let LiKafkaConsumerImpl to handleInvalidOffsetException (i.e. auto reset to latest/earliest/closest based on config) and also throw the exception so that XC can know an exception happened and invoke the user callback. The end goal is to move the large msg related logic to XC layer so that we only use the raw kafka consumer underlying so everything is handled at XC layer. 

This is just a temp solution since if we handleInvalidOffsetException in XC layer, we'll mess up the ConsumerRecordsProcessor state (large msg hwm, safe offset etc), so we'll let LiKafkaConsumerImpl handle it as it is now, but force it to throw exception.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -72,6 +73,12 @@
       "full or the incomplete message has expired. The consumer will throw a LargeMessageDroppedException if this " +
       "configuration is set to true. Otherwise the consumer will drop the message silently.";
 
+  private static final String EXCEPTION_ON_INVALID_OFFSET_DOC = "If true, even when auto.offset.reset is set to " +
+      "earliest/latest/liclosest, InvalidOffsetException will be thrown to user upon calling to poll. This is to allow " +
+      "XinfraClients to properly invoke user callback upon InvalidOffsetException while handling the reset and corresponding " +
```
xiowu0, line 78: should we avoid mention "XinfraClients" in the doc, since "XinfraClients" is not open-sourced yet?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerConfig.java
```diff
@@ -33,6 +33,7 @@
   public static final String MESSAGE_ASSEMBLER_EXPIRATION_OFFSET_GAP_CONFIG = "message.assembler.expiration.offset.gap";
   public static final String MAX_TRACKED_MESSAGES_PER_PARTITION_CONFIG = "max.tracked.messages.per.partition";
   public static final String EXCEPTION_ON_MESSAGE_DROPPED_CONFIG = "exception.on.message.dropped";
+  public static final String EXCEPTION_ON_INVALID_OFFSET_CONFIG = "exception.on.invalid.offset";
```
xiowu0, line 36: rename to : 
exception.on.invalid.offset.reset?
an exception is thrown anyway if offset reset strategy is none.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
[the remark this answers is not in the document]
navina: mm ok. So, on exception, xinfra client invokes user callback? dumb question: what user callback are we referring to here?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
```diff
@@ -319,6 +322,9 @@ public void unsubscribe() {
...
+          ioe = oe;
```
xiowu0, line 326: The exception can be thrown right here.  If we see OffsetOutOfRangeException here, rawRecords is guaranteed to be an empty map. So we don't need to wait until records are processed before throwing the exception.

navina: +1

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/consumer/LiKafkaConsumerImpl.java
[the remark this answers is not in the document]
kehuum: 1. Yes
2. We don't have any concrete usage yet, but likely pause/re-assign etc.

PR #184, opened 2020-08-08
Add enableRecordHeader config in producer

bgeng7:
This PR adds `enableRecordHeader` config in producer. When it is true, producer will add a new field in the record header and skip segment serialization in MsgSplitter. The type(version) field in LMHeaderValue will be set to V3 as well, which means we will only use record header based LM support. 
By detecting the type(version) field in LMHeaderValue, consumer  can choose if it should use segment deserialization in MsgAssembler. So we do not need to add this config in consumer side.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
@@ -51,6 +51,7 @@ public class LiKafkaProducerConfig extends AbstractConfig {
   public static final String MAX_REQUEST_SIZE_CONFIG = ProducerConfig.MAX_REQUEST_SIZE_CONFIG;
   public static final String LARGE_MESSAGE_SEGMENT_WRAPPING_REQUIRED_CONFIG =
       "li.large.message.segment.wrapping.required";
+  public static final String ENABLE_RECORD_HEADER_CONFIG = "li.record.header.enable";
 
   public static final String LARGE_MESSAGE_ENABLED_DOC = "Configure the producer to support large messages or not. " +
       "If large message is enabled, the producer will split the messages whose size is greater than " +
@@ -98,6 +99,8 @@ public class LiKafkaProducerConfig extends AbstractConfig {
       "and thus not split into multiple messages. This configuration does not have any effect if large message is " +
       "not enabled.";
 
+  public static final String ENABLE_RECORD_HEADER_DOC = "If true, we will use record header based large message support and stop using segmentSerializer.";
+
   static {
     // TODO: Add a default metadata service client class.
     CONFIG = new ConfigDef()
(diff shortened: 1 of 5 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerImpl.java
```diff
@@ -301,6 +304,18 @@ private LiKafkaProducerImpl(LiKafkaProducerConfig configs,
         }
       }
 
+      // TODO: discuss with navina if adding enableRecordHeader in header to pass parameter is a good choice.
+      // It may be a little strange to control if using record headers by testing if the record header has such a field.
+      if (_enableRecordHeader) {
```
navina, line 309: This is a producer instance-level configuration and hence, doesn't need to be inside every record send. 
You can check for the config in the producer constructor and pass the value as a boolean argument to the `MessageSplitterImpl` constructor.

bgeng7: Yeah, it should be a producer instance-level configuration.  I will update to put the config in the constructor.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/Constants.java
```diff
@@ -9,6 +9,7 @@
   public static final String TIMESTAMP_HEADER = "_t";
   public static final String LARGE_MESSAGE_HEADER = "_lm";
   public static final String SAFE_OFFSET_HEADER = "_so";
+  public static final String SHOULD_USE_HEADER = "_sh";
```
navina, line 12: This is no longer needed.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
```diff
@@ -120,6 +103,32 @@ public static int decodeInt(byte[] input, int pos) {
    * @return a decoded long
    */
   public static long decodeLong(byte[] input, int pos) {
+    sanityCheck(input, pos, LONG_SIZE);
```
navina, line 106: Not sure I follow the changes in this class. Can you please explain?  Is it just refactoring code ?

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/Constants.java
bgeng7: Removed.

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/utils/PrimitiveEncoderDecoder.java
bgeng7: Yes, it is just refactoring. In the first version, I add a new encode/decode method for boolean type for adding header. When I was doing that, I found there were some duplicate codes for sanity check. So I extracted them. After removing extra header, I also removed unused encodeBoolean method but left the sanity check method.

navina: Ok

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaProducerConfig.java
```diff
@@ -98,6 +99,8 @@
       "and thus not split into multiple messages. This configuration does not have any effect if large message is " +
       "not enabled.";
 
+  public static final String ENABLE_RECORD_HEADER_DOC = "If true, we will use record header based large message support and stop using segmentSerializer.";
```
xiowu0, line 102: Is this config only used for large message support?  If so, could you rename the config to include "large.message" in the config name?  If not, could you update the doc to mention there could be other use of it

bgeng7: It will be used for large message support and other kinds of serialization(e.g. encryption or compression). I have updated the comments

PR #186, opened 2020-08-14
add test and fix a bug of using  payload.array()

bgeng7:
The previous (pr)[https://github.com/linkedin/li-apache-kafka-clients/pull/184] introduce a bug of   wrongly using payload.array() instead of using segment.payloadArray().
This pr is to fix that problem and add a new integration test for that config

### DIFF
--- a/integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
+++ b/integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
@@ -7,6 +7,7 @@ package com.linkedin.kafka.clients.largemessage;
 import com.linkedin.kafka.clients.common.LargeMessageHeaderValue;
 import com.linkedin.kafka.clients.consumer.LiKafkaConsumer;
 import com.linkedin.kafka.clients.producer.LiKafkaProducer;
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/com/linkedin/kafka/clients/largemessage/LargeMessageIntegrationTest.java
```diff
@@ -175,6 +176,87 @@ public void testLargeMessage() throws Exception {
...
+        consumptionStarted = true;
+      }
+      for (ConsumerRecord<String, String> consumerRecord : records) {
+        // Verify headers
+        Long eventTimestamp = LiKafkaClientsUtils.fetchTimestampHeader(consumerRecord.headers());
+        assertNotNull(eventTimestamp);
+        assertTrue(eventTimestamp >= startTime && eventTimestamp <= System.currentTimeMillis());
+        LargeMessageHeaderValue largeMessageHeaderValue = LiKafkaClientsUtils.fetchLargeMessageHeader(consumerRecord.headers());
+        assertEquals(largeMessageHeaderValue.getSegmentNumber(), -1);
+        assertEquals(largeMessageHeaderValue.getNumberOfSegments(), 6);
```
xiowu0, line 246: nit:  the number of segments can be calculated based on the record size and max message size.

PR #189, opened 2020-09-03
Fix race condition in LiKafkaInstrumentedProducerImpl

kehuum:
Fix race condition in LiKafkaInstrumentedProducerImpl when two threads calling close in send callback.

Problem:
BMM has a use case of calling producer.close in send callback when send fails, with current code, there could still be race conditions given the following sequence:

Both Thread 1 and Thread 2 are calling same producer.send(tp, cbk) and fails, and tries to call producer.close in the cbk:

Thread 1: Hold closeLock, (release Read lock,) wait on Write lock
Thread 2: Hold Read lock, wait on closeLock.

Thus deadlock.

Fix:
Ensure same ordering of acquiring/releasing locks when proceedClose(), i.e. first release all Read locks held by current thread, then grab Write lock to proceed closing.

### DIFF
--- a/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
+++ b/li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
@@ -461,11 +461,11 @@ public class LiKafkaInstrumentedProducerImpl<K, V> implements DelegatingProducer
     }
 
     // release read lock held by current thread if any
-    // BMM has use cases of closing the producer in send callback if send fails, we need to ensure
+    // There are use cases of closing the producer in send callback if send fails, we need to ensure
     // the same ordering of acquiring/releasing locks, in this case:
     // 1. send will get the Read lock
     // 2. close will first release all Read locks; then grab the Write lock
-    int holds = delegateLock.getReadHoldCount();
+    final int holds = delegateLock.getReadHoldCount();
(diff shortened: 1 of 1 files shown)


li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -460,36 +459,39 @@ private boolean proceedClosing() {
...
-      ReentrantReadWriteLock.WriteLock writeLock = delegateLock.writeLock();
-      if (holds > 0) { //do we own a read lock ?
-        for (int i = 0; i < holds; i++) {
-          readLock.unlock();
-        }
-        //at this point we no longer hold a read lock, but any number of other
-        //readers/writers may slip past us
+
+    // release read lock held by current thread if any
+    // BMM has use cases of closing the producer in send callback if send fails, we need to ensure
```
radai-rosenblatt, line 464: nit - not specific to bmm. closing from a callback is sadly a kafka feature

radai-rosenblatt (approved):

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
jonlee2: Is there a way to disable that? Of course I am not suggesting we do this in this PR.

xiowu0 (approved):

PR #192, opened 2021-02-17
Fix Kafka flush() contract different in vanilla client v/s li-apache-kafka-clients

fluffywei:
This request aims to fixing li-apache-kafka-clients producer flush() contract different in vanilla client.

- Removed verifyOpen() from flush() and flush(long timeout, TimeUnit timeUnit).
- Peeled InvocationTargetException and re-throw vanilla client exception.
- Added flush() integration test.

### DIFF
--- a/integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
+++ b/integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
@@ -197,7 +197,7 @@ public class LiKafkaInstrumentedProducerIntegrationTest extends AbstractKafkaCli
   }
 
   @Test
-  public void testProducerFlush() throws Exception {
+  public void testProducerFlushAfterClose() throws Exception {
     String topic = "testCloseFromProduceCallbackOnSenderThread";
     createTopic(topic, 1);
 
@@ -222,9 +222,9 @@ public class LiKafkaInstrumentedProducerIntegrationTest extends AbstractKafkaCli
     random.nextBytes(value);
     ProducerRecord<byte[], byte[]> record = new ProducerRecord<>(topic, key, value);
 
-    producer.send(record).get();
-    producer.flush(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
+    producer.send(record);
     producer.close(Duration.ofSeconds(30));
+    producer.flush(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
   }
 
   private void createTopic(String topicName, int numPartitions) throws Exception {
--- a/integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
+++ b/integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
@@ -31,6 +31,7 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.errors.RecordTooLargeException;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.testng.Assert;
 import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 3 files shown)


fluffywei:
@kehuum

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
```diff
@@ -196,6 +196,37 @@ public void uncaughtException(Thread t, Throwable e) {
     Assert.assertTrue(root instanceof RecordTooLargeException, root.getMessage());
   }
 
+  @Test
+  public void testProducerFlush() throws Exception {
```
kehuum, line 200: rename to testProducerFlushAfterClose

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
```diff
@@ -196,6 +196,37 @@ public void uncaughtException(Thread t, Throwable e) {
...
+        10
+    );
+
+    byte[] key = new byte[500];
+    byte[] value = new byte[500];
+    random.nextBytes(key);
+    random.nextBytes(value);
+    ProducerRecord<byte[], byte[]> record = new ProducerRecord<>(topic, key, value);
+
+    producer.send(record).get();
```
kehuum, line 225: probably don't need get() to allow flush doing the work, get will wait for the record to be acked.

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
```diff
@@ -196,6 +196,37 @@ public void uncaughtException(Thread t, Throwable e) {
...
+    producer.flush(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
+    producer.close(Duration.ofSeconds(30));
```
kehuum, line 227: close before flush

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -355,16 +353,20 @@ public void flush() {
...
         try {
           boundedFlushMethod.invoke(delegate, timeout, timeUnit);
-        } catch (IllegalAccessException | InvocationTargetException e) {
-          throw new IllegalStateException("failed to invoke the bounded flush method", e);
+        } catch (IllegalAccessException illegalAccessException) {
+          throw new IllegalStateException("Failed to invoke the bounded flush method", illegalAccessException.getCause());
+        } catch (InvocationTargetException invocationTargetException) {
+          try {
+            throw (Exception) invocationTargetException.getCause();
+          } catch (Exception exception) {
```
xiowu0, line 367: this could catch the exception thrown in line 366. 
One way to fix is to check if the "cause" is an instance of runtime exception
if the cause is a runtime exception, cast the exception to runtime exception and rethrow the exception.
if the cause is not a runtime exception, wrap the exception with the IllegalStateException as before.

fluffywei: Hi Wesley, I have resolved this part according to your comments. Feel free to ping me if further modification is required. Thank you!

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -355,17 +353,19 @@ public void flush() {
...
       boolean useSeparateThreadForFlush = false;
       if (boundedFlushMethod != null) {
         try {
           boundedFlushMethod.invoke(delegate, timeout, timeUnit);
-        } catch (IllegalAccessException | InvocationTargetException e) {
-          throw new IllegalStateException("failed to invoke the bounded flush method", e);
-        }
+        } catch (Exception e) {
+            if (e.getCause() instanceof RuntimeException) {
+              throw new RuntimeException("Failed to flush due to RuntimeException", e.getCause());
```
xiowu0, line 364: this should be :
throw (RuntimeException) e.getCause()

li-apache-kafka-clients/src/main/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerImpl.java
```diff
@@ -355,17 +353,19 @@ public void flush() {
...
     delegateLock.readLock().lock();
     try {
       boolean useSeparateThreadForFlush = false;
       if (boundedFlushMethod != null) {
         try {
           boundedFlushMethod.invoke(delegate, timeout, timeUnit);
-        } catch (IllegalAccessException | InvocationTargetException e) {
-          throw new IllegalStateException("failed to invoke the bounded flush method", e);
-        }
+        } catch (Exception e) {
```
xiowu0, line 362: catch InvocationTargetException here

Separate IllegalAccessException from InvocationTargetException here,  we only need to find the cause for InvocationTargetException. 

keep the handling of IllegalAccessException behavior unchanged as before.

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
```diff
@@ -222,14 +223,43 @@ public void testProducerFlushAfterClose() throws Exception {
     random.nextBytes(value);
     ProducerRecord<byte[], byte[]> record = new ProducerRecord<>(topic, key, value);
 
+    producer.send(record);
+    producer.close(Duration.ofSeconds(0));
+    producer.flush(0, TimeUnit.MILLISECONDS);
+  }
+
+  @Test
+  public void testProducerFlushWithBrokerKilled() throws Exception {
```
xiowu0, line 232: rename the test case: testProducerFlushWithTimeoutException ?

xiowu0 (approved):

integration-tests/src/test/java/com/linkedin/kafka/clients/producer/LiKafkaInstrumentedProducerIntegrationTest.java
```diff
@@ -196,6 +198,70 @@ public void uncaughtException(Thread t, Throwable e) {
     Assert.assertTrue(root instanceof RecordTooLargeException, root.getMessage());
   }
 
+  @Test
+  public void testProducerFlushAfterClose() throws Exception {
```
vmaheshw, line 202: Have you tested what happens when the flush(timeout) is called when the close() is in progress and has not completed?

What does it return if timeout is less than the time it takes the close() to complete V/s timeout is more than the time it takes the close() to complete? In second case, will it return immediately after close() is completed or wait for the timeout

xiowu0: It is overkill to have unit tests for all these scenarios given the fact that we might need to put extra wait time with different timeout and synchronization between threads.  we try to avoid "wait for timeout" in the unit tests as much as possible.  These two unit tests have covered the two new features:
1) allow timeoutexception to be propagated to users
2) allow flush to be called after close.

if you want to understand the behavior of flush when doing concurrent close,  we can write some local tests, and let you know the behavior of the following case:
1) a) close with timeout e.g., 10 seconds  in thread 1, b) flush with timeout e.g, 5 seconds in thread 2, see if flush hits a timeout.
2) a) close with timeout e.g. 5 seconds in thread 1, b) flush with timeout e.g., 10 seconds in thread 1, measure the time when flush finished.

vmaheshw: local tests will work. Thanks. I will wait to hear back about the behavior.

xiowu0: The behavior is as follows:
1) case 1 -> flush time out
2) case 2 -> flush completed without a timeout.

example test code:

case 1

// close 
new Thread(() -> producer.close(Duration.ofSeconds(3))).start();
// wait close to in progress. 
Thread.sleep(1000);
producer.flush(1, TimeUnit.SECONDS);  ==> this timed out as expected.

case 2

// close 
new Thread(() -> producer.close(Duration.ofSeconds(3))).start();
// wait close to in progress. 
Thread.sleep(1000);
producer.flush(4, TimeUnit.SECONDS);  ==> this completes sucessfully without timeout.

